PR #89, opened 2026-04-15
Add stateless field to registry server types

### LINKED ISSUE #88: Add stateless field to registry server types
## Summary

ToolHive's `--stateless` flag (stacklok/toolhive#4515) lets the proxy handle stateless streamable-HTTP servers correctly — servers like Spring AI's `WebMvcStatelessServerTransport` that accept POST only and don't serve SSE on GET. The flag works, but it's a CLI-only property. The registry schema has no way to express it.

This means every user running a stateless server through a registry must know to pass `--stateless` manually. The registry should be the single source of truth for how to run a server. Transport type, proxy port, and OAuth config are all declarable in the registry today. Stateless mode is the same kind of property.

## Proposed change

Add a `Stateless bool` field to `BaseServerMetadata`, `ServerExtensions`, and the JSON schemas (legacy registry and publisher-provided). Update the bidirectional converters to preserve the field through round-trips.

This is the type-system half. A follow-up PR against stacklok/toolhive will read this field in `buildRunnerConfig` so that `thv run <server>` applies stateless mode automatically when the registry declares it.

## Downstream impact

The `ServerMetadata` interface gains `GetStateless() bool`, following the pattern of the 14 existing `BaseServerMetadata` getters. No downstream repo implements `ServerMetadata` directly. The field uses `omitempty` and defaults to `false`, so existing registry data is unaffected.

gkatz2:
## Summary

- Registry entries cannot declare that a server is stateless (POST-only, no SSE/GET). Users must pass `--stateless` manually on every `thv run` invocation. Adding the field to the type system lets registries declare this property so it applies automatically.
- Adds `Stateless bool` to `BaseServerMetadata`, `ServerExtensions`, both legacy registry schema definitions (`server` and `remote_server`), and the publisher-provided schema. Updates bidirectional converters to preserve the field through round-trips.

Fixes #88

## Design notes

**Why `BaseServerMetadata` (shared) rather than `RemoteServerMetadata` only?** Stateless mode applies to both remote servers and container workloads — stacklok/toolhive#4515 explicitly supports both. Placing the field on `BaseServerMetadata` alongside `Transport` is consistent with how other transport-level properties are modeled. The trade-off is that `ImageMetadata` inherits a field that's irrelevant for stdio containers, but this mirrors existing fields like `TargetPort` and `ProxyPort` that only apply to HTTP transports.

**Interface addition.** `GetStateless()` is added to `ServerMetadata`, following the established pattern for all 14 existing `BaseServerMetadata` getters. Unlike `ProxyPort` (which predates the interface and is duplicated on both concrete types), `Stateless` follows the dominant pattern. No downstream repo implements `ServerMetadata` directly — all consumers use the concrete types, which inherit the getter through embedding.

## Test plan

- [x] Unit tests: converter round-trip for stateless=true on both remote and image servers
- [x] Unit tests: YAML/JSON marshal/unmarshal preserves the field
- [x] Unit tests: nil receiver returns false
- [x] Unit tests: `GetStateless()` through `ServerMetadata` interface
- [x] Existing "all fields" round-trip test updated to include stateless
- [x] `task lint-fix` — 0 issues
- [x] `task test` — all pass with race detection
- [x] `task license-check` — all headers present

### DIFF
--- a/registry/converters/converters_test.go
+++ b/registry/converters/converters_test.go
@@ -882,6 +882,76 @@ func TestRoundTrip_RemoteServerMetadata(t *testing.T) {
 	assert.Len(t, result.ToolDefinitions, len(original.ToolDefinitions))
 }
 
+func TestRoundTrip_StatelessField(t *testing.T) {
+	t.Parallel()
+
+	t.Run("remote server with stateless=true", func(t *testing.T) {
+		t.Parallel()
+
+		original := &registry.RemoteServerMetadata{
+			BaseServerMetadata: registry.BaseServerMetadata{
+				Description: "Stateless remote server",
+				Transport:   "streamable-http",
+				Status:      "active",
+				Tier:        "Official",
+				Tools:       []string{"tool1"},
+				Tags:        []string{"stateless"},
+				Stateless:   true,
+			},
+			URL: "https://api.example.com/mcp",
+		}
+
+		serverJSON, err := RemoteServerMetadataToServerJSON("stateless-remote", original)
+		require.NoError(t, err)
+
+		result, err := ServerJSONToRemoteServerMetadata(serverJSON)
+		require.NoError(t, err)
+
+		assert.True(t, result.Stateless, "Stateless field should be preserved through round-trip")
+		assert.True(t, result.GetStateless(), "GetStateless() should return true")
+	})
+
+	t.Run("remote server with stateless=false (default)", func(t *testing.T) {
+		t.Parallel()
+
+		original := createTestRemoteServerMetadata()
+		assert.False(t, original.Stateless, "test fixture should default to false")
+
+		serverJSON, err := RemoteServerMetadataToServerJSON("stateful-remote", original)
+		require.NoError(t, err)
+
+		result, err := ServerJSONToRemoteServerMetadata(serverJSON)
+		require.NoError(t, err)
+
+		assert.False(t, result.Stateless, "Stateless field should remain false")
+	})
+
+	t.Run("image server with stateless=true", func(t *testing.T) {
+		t.Parallel()
+
+		original := &registry.ImageMetadata{
+			BaseServerMetadata: registry.BaseServerMetadata{
+				Description: "Stateless container server",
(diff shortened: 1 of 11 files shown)


jerm-dro:
Hey @gkatz2, thanks for the thorough write-up and for following up on the `--stateless` work!

  I've been thinking about this, and I'm hesitant to promote stateless to a first-class registry property. My concern is that stateless mode is really a runtime behavior of the server. Whether it accepts only POST and doesn't serve SSE on GET feels like something the proxy should be able to detect automatically rather than requiring explicit declaration.

For example, when thv run starts proxying a server, it could probe the server's behavior (e.g., attempt a GET and observe a 405 or lack of SSE response) and automatically enable stateless mode. This would mean:
  1. No `--stateless` flag needed on `thv run`
  2. No registry metadata needed
  3. It "just works" for any stateless server without users needing to know about the distinction

Before we add this to the type system, could you explore whether automatic detection at runtime is feasible for your use case?

gkatz2:
Thanks for the thoughtful feedback, @jerm-dro!

Auto-detection is a good idea and the spec is clear (MUST return 405 on GET for stateless servers), so it would work for compliant servers. Two complications make it insufficient as the sole mechanism: some frameworks don't return 405 by default when a route is only registered for POST ([e.g., Express.js returns 404](https://github.com/expressjs/express/issues/2414)), and intermediary proxies can return their own responses before the request reaches the MCP server. In those cases you'd still need an explicit override.

The registry field covers all cases declaratively. The operator knows their server is stateless and declares it, same as transport type or OAuth config. Even if auto-detection lands later, the registry field gives operators a reliable override when detection gets it wrong.

What do you think?

jerm-dro:
Thanks for the context @gkatz2. I had not considered the possibility that auto-detection would fail. For the Express.js example, would it be possible to fix address the root cause upstream rather than working around it here? Is that much effort?

I'm curious what @rdimitrov thinks about this. What's the cost/resistance to including this workaround?

stacklokbot (approved):

PR #111, opened 2026-05-18
Add shared postgres client package

reyortiz3:
## Summary

Introduces a `postgres/` package mirroring `redis/`: a validated `Config`, a `NewPool` factory that wraps `pgxpool`, and a dynamic-authentication dispatcher with AWS RDS IAM as the first backend.

This is the consolidation point for PostgreSQL connection plumbing currently duplicated in [toolhive-registry-server](https://github.com/stacklok/toolhive-registry-server) (`internal/app/storage/database_factory.go` + `internal/app/storage/auth/`). Additional consumers planned. **Schema concerns — migrations, sqlc bindings, application-specific type codecs — stay in each caller.**

## Public surface

- `postgres.Config` — pool tuning, app + migration user, `DynamicAuth` block. Implements `slog.LogValuer` to redact credentials.
- `postgres.NewPool(ctx, *Config, ...Option)` — auto-installs a `BeforeConnect` dynamic-auth hook when `DynamicAuth` is set.
- Options: `WithAfterConnect` (for custom type codecs), `WithBeforeConnect` (escape hatch), `WithLogger`.
- `postgres.NewAuthToken` / `postgres.NewDynamicAuthFunc` — token / hook constructors for callers that need them outside `NewPool` (e.g. golang-migrate's one-shot connection).
- `postgres.StartPoolStatsLogger` — opt-in periodic pool-stats emitter.
- `postgres.DynamicAuthAWSRDSIAM` — `Region: "detect"` discovers region via IMDS; static regions are used as-is.

## Design notes

- **Single package**, not a sub-package per backend. Adding Vault/GCP-IAM later means a new `*backend*.go` file and a new dispatcher branch — no API churn at the call sites.
- **No mutual-exclusion of `Password` and `DynamicAuth`** at the library boundary. Some callers want a local-dev static fallback. Mutual exclusion belongs in the consumer's config-validation layer (per registry-server's `.claude/rules/secrets.md` §4).
- **Password is a plain string.** Secret-file reads, env-var overrides, and pgpass fallback all stay in the consumer's config loader; this package never opens a file.
- **`pgxpool.NewWithConfig` is lazy.** `NewPool` returns without verifying connectivity; dial errors surface on first acquire, matching pgxpool's design.

## Test plan

- [x] `task lint` — 0 issues
- [x] `task test` — all packages pass under `-race`
- [x] Coverage 82.5% (above the 70% bar in CLAUDE.md). Uncovered code is the AWS SDK call sites that require real credentials to exercise.


rdimitrov (dismissed):

postgres/config.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	var userInfo *url.Userinfo
+	if password != "" {
+		userInfo = url.UserPassword(user, password)
+	} else {
+		userInfo = url.User(user)
+	}
+
+	return fmt.Sprintf(
```
jhrozek, line 217: **[HIGH] `Host` and `Database` are interpolated raw into the DSN.**

`url.UserPassword` correctly percent-encodes the userinfo segment, but `c.Host` and `c.Database` flow into `fmt.Sprintf` unescaped, and `Validate()` only checks emptiness. A `Database` containing `?` silently overrides `sslmode`; a `Host` containing `@` is parsed by `pgxpool.ParseConfig`/`url.Parse` as additional userinfo (e.g. `Host: "good.rds@evil.example.com"` connects to `evil.example.com` carrying credentials for the `good.rds` host).

Recommendation: build `pgxpool.Config` structurally rather than formatting a URL string — this removes the parsing path entirely:

```go
pgCfg, err := pgx.ParseConfig("")
if err != nil { return nil, err }
pgCfg.Host     = c.Host
pgCfg.Port     = uint16(c.Port)
pgCfg.User     = c.User
pgCfg.Password = c.Password
pgCfg.Database = c.Database
// then pgxpool.NewWithConfig with a pgxpool.Config that wraps pgCfg
```

Alternative: use `url.URL{Scheme:"postgres", User:userInfo, Host:net.JoinHostPort(c.Host, strconv.Itoa(c.Port)), Path:"/"+c.Database, RawQuery:"sslmode="+sslMode}.String()` and let the stdlib do the escaping. Or add character-class validation for `Host`/`Database` in `Validate()`.

postgres/config.go
```diff
@@ -0,0 +1,201 @@
...
+	// time via NewPool's automatically-installed BeforeConnect hook.
+	DynamicAuth *DynamicAuthConfig
+
+	// MaxOpenConns sets the upper bound on open connections in the pool. When
+	// zero, pgxpool's default is used.
+	MaxOpenConns int32
+
+	// MaxIdleConns sets the minimum number of idle connections the pool tries
+	// to maintain. When zero, pgxpool's default is used.
+	MaxIdleConns int32
```
jhrozek: **[HIGH] `MaxIdleConns` is wired to pgxpool's `MinConns` — opposite semantics.**

In `database/sql`, `MaxIdleConns` is a *ceiling* on connections sitting idle. In pgxpool, `MinConns` is a *floor* — the pool actively maintains that many open connections even when the application is quiet. Setting `MaxIdleConns: 50` on this `Config` will keep 50 permanent connections open against the database, the opposite of what a `database/sql`-trained developer expects.

The doc comment is accurate, but the field name will override the doc in most readers' heads. This is a public API on a shared library — renaming after stabilization is a breaking change. Suggest renaming to `MinConns` (or `MinIdleConns`) now while the package is still Alpha.

```suggestion
	// MinConns is the minimum number of connections pgxpool actively
	// maintains in the pool. When zero, pgxpool's default is used.
	MinConns int32
```

(Don't forget to update `applyPoolTuning` in `pool.go` to read the renamed field.)

postgres/pool.go
```diff
@@ -0,0 +1,158 @@
...
+
+	beforeConnect := o.beforeConnect
+	if beforeConnect == nil && cfg.DynamicAuth != nil {
+		beforeConnect, err = NewDynamicAuthFunc(ctx, cfg, cfg.User)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if beforeConnect != nil {
+		poolConfig.BeforeConnect = beforeConnect
```
jhrozek, line 106: **[MEDIUM] `WithBeforeConnect` silently replaces the auto-installed dynamic-auth hook.**

When `cfg.DynamicAuth` is set and the caller also passes `WithBeforeConnect(metricsHook)`, the user hook silently wins — the auth hook is never built. There's no error, no warning, no log. The doc on `WithBeforeConnect` calls this out, but documented footguns are still footguns: the failure mode here is "production tokens expire ~15 min after deploy."

Recommend composing by default — call the auth hook first, then the user hook — and reserving an explicit `WithReplaceBeforeConnect` for the override case. Sketch:

```go
var authHook BeforeConnectFn
if cfg.DynamicAuth != nil {
    authHook, err = NewDynamicAuthFunc(ctx, cfg, cfg.User)
    if err != nil { return nil, err }
}
switch {
case authHook != nil && o.beforeConnect != nil:
    poolConfig.BeforeConnect = func(ctx context.Context, c *pgx.ConnConfig) error {
        if err := authHook(ctx, c); err != nil { return err }
        return o.beforeConnect(ctx, c)
    }
case authHook != nil:
    poolConfig.BeforeConnect = authHook
case o.beforeConnect != nil:
    poolConfig.BeforeConnect = o.beforeConnect
}
```

postgres/config.go
```diff
@@ -0,0 +1,201 @@
...
+	"log/slog"
+	"net/url"
+	"time"
+)
+
+// DefaultSSLMode is applied by BuildConnectionStringWithAuth when Config.SSLMode
+// is empty. "require" is the safe production default: encryption is mandatory
+// but the server certificate is not validated against a CA. Callers that need
+// stricter behavior should set SSLMode to "verify-ca" or "verify-full".
+const DefaultSSLMode = "require"
```
jhrozek, line 29: **[MEDIUM] `DefaultSSLMode = "require"` does not verify the server certificate.**

`sslmode=require` defends against passive eavesdropping but not active MITM — a network attacker that can intercept TCP can present any certificate and the connection succeeds. The doc comment is accurate about the semantics, but calling it "the safe production default" overstates the protection. AWS RDS publishes a CA bundle precisely so callers can use `verify-full` against RDS's PKI.

For a shared library, recommend changing the default to `"verify-full"` and documenting the downgrade path (dev environments with self-signed certs need to opt down explicitly). At minimum, reword the comment to drop "safe production default."

postgres/config.go
```diff
@@ -0,0 +1,201 @@
...
+		return errors.New("user is required")
+	}
+	if c.Database == "" {
+		return errors.New("database is required")
+	}
+	if c.DynamicAuth != nil {
+		if c.DynamicAuth.AWSRDSIAM == nil {
+			return errors.New("dynamicAuth is set but no supported auth method (e.g., awsRdsIam) is configured")
+		}
+		if c.DynamicAuth.AWSRDSIAM != nil && c.DynamicAuth.AWSRDSIAM.Region == "" {
```
jhrozek: **[LOW] Dead nil-check.** After the early return on line 108 (`AWSRDSIAM == nil`), the `c.DynamicAuth.AWSRDSIAM != nil` part of this conjunction is always true. `staticcheck` will flag this as SA4031.

```suggestion
		if c.DynamicAuth.AWSRDSIAM.Region == "" {
```

postgres/doc.go
```diff
@@ -0,0 +1,81 @@
...
+level until its context is cancelled.
+
+# Secrets Handling
+
+This package treats Password and MigrationPassword as already-resolved
+strings. File-based secret loading, environment-variable overrides, and
+pgpass fallback all live in the caller's configuration layer. Config
+implements slog.LogValuer to redact credentials when logged.
+*/
+package postgres
```
jhrozek, line 100: **[LOW] Add a stability marker + update the top-level package tables.**

This package is missing a stability indicator. Per the repo's graduation conventions in `README.md` and `CLAUDE.md`, new packages should declare their track. Suggest:

1. Add `// Stability: Alpha` at the top of this doc comment (or as a separate top-line comment above `package postgres`).
2. Add a row to `README.md`'s **Available Packages** table:
   `| postgres | Alpha | PostgreSQL connection pool with optional AWS RDS IAM dynamic auth |`
3. Add a row to `CLAUDE.md`'s **Current Packages** table:
   `| postgres | PostgreSQL connection pool with optional AWS RDS IAM dynamic auth (Alpha) |`

(Side note: `redis/` is also missing from both tables — likely a separate cleanup, not in scope for this PR.)

jhrozek (commented):
Batched review of the postgres/ package. Six findings — two HIGH (DSN injection / pool tuning naming), two MEDIUM (auth-hook composition / SSL default), two LOW (dead nil-check / stability marker). None blocking, but recommend addressing the HIGH and MEDIUM before tagging a release. Test coverage at 82.5% and lint clean.

reyortiz3:
Pushed 4 commits addressing the HIGH and LOW findings. The two MEDIUMs are still open — counter-proposals below.

## Addressed

| Finding | Commit | Notes |
|---|---|---|
| [HIGH] DSN injection via `Host`/`Database` | d6c19e7 | Verified both attacks empirically — `url.Parse` on `good.rds@evil:5432` resolves to `evil:5432`, and a `?` in `Database` produces a second query section that pgx resolves in caller-hostile order. Went with a hybrid fix: structural override of `Host`/`Port`/`User`/`Password`/`Database` onto a `pgxpool.Config` bootstrapped from a minimal `sslmode=X` DSN (so pgx still translates SSL mode → `*tls.Config`), plus character-class validation in `Validate()` so the `ConnectionString` methods are also safe for migration-tool callers. Also tightened the `Port` check to `1..65535` to give the `uint16` conversion a real invariant. |
| [HIGH] `MaxIdleConns` → `MinConns` | ba45e27 | Renamed plus a doc comment that calls out the `database/sql` contrast for readers carrying that mental model. |
| [LOW] Dead nil-check | de3e617 | Collapsed as suggested. |
| [LOW] Stability marker + tables | 13ca610 | Added `// Stability: Alpha`, plus rows in `README.md` and `CLAUDE.md`. Agree `redis/` is missing from both — happy to handle as a separate follow-up. |

## Open — pushing back on both MEDIUMs

**[MEDIUM] `WithBeforeConnect` silently replaces dynamic-auth hook.** Agree the silent override is a footgun, but compose-by-default has its own footgun: both hooks write to `conn.Password`, and which one wins depends on order. The "want to layer metrics on top of auth" case is rare; so is "replace the auth hook entirely." I'd rather **error loudly** when both `cfg.DynamicAuth != nil` and `WithBeforeConnect` are set — force the caller to pick. Callers who really mean to replace the auth hook leave `cfg.DynamicAuth = nil` and pass their own. No silent failure, no surprising composition order. WDYT?

**[MEDIUM] `DefaultSSLMode = "require"` doesn't verify cert.** Agree the comment overstates protection. Less convinced about flipping the default to `verify-full`:
- `verify-full` requires a working CA chain. Cloud RDS works (Amazon publishes a bundle), but self-hosted Postgres with a self-signed cert — common in dev/staging — breaks on first run.
- registry-server today uses `require`. A default flip would silently change behavior on bump unless we also pin `SSLMode: "require"` at the extraction step.

Proposal: keep `require` as the library default (least surprise on extraction), drop "safe production default" from the doc, and add a paragraph in `doc.go` recommending `verify-full` for production cloud Postgres with CA bundles. Security guidance in the docs, not in a default that breaks dev environments. WDYT?

reyortiz3:
Pushed both MEDIUM fixes — but didn't take the reviewer's suggested shape on either; took the alternatives from my earlier reply instead.

| Finding | Commit | Approach |
|---|---|---|
| [MEDIUM] `WithBeforeConnect` silently replaces dynamic-auth | c9327c4 | Reject the combination at construction time rather than compose. NewPool returns an error when `cfg.DynamicAuth != nil` AND `WithBeforeConnect` is also set. Callers that genuinely want both compose the hooks themselves via `NewDynamicAuthFunc` + custom hook, then pass the composition with `cfg.DynamicAuth = nil`. No ambiguous default order; no silent token expiry. |
| [MEDIUM] `DefaultSSLMode = "require"` overstates protection | e2caa83 | Kept the default value at `"require"` (flipping to `verify-full` would silently break self-signed / dev environments on first run). Dropped the "safe production default" wording from the const comment, and added a TLS and SSL Mode section in `doc.go` that explains the tradeoff and points production cloud-Postgres deployments at `verify-full` with a CA bundle. |

Happy to revisit either if the reasoning doesn't land — open to a compose-by-default `WithBeforeConnect` if you'd rather not error, and to bumping the default to `verify-full` if breaking dev environments is acceptable as part of the security stance.

jhrozek (approved):

postgres/config.go
reyortiz3: addressed, see comment below

postgres/config.go
reyortiz3: addressed, see comment below

postgres/doc.go
reyortiz3: addressed, see comment below

postgres/pool.go
reyortiz3: addressed, see comment below

postgres/config.go
reyortiz3: addressed, see comment below

postgres/config.go
reyortiz3: addressed, see comment below

PR #133, opened 2026-06-15
oci: extract artifact-agnostic primitives into oci/artifact (Phase 0, THV-0077)

### LINKED ISSUE #130: [plugins] Phase 0: extract artifact-agnostic OCI primitives (THV-0077)
**Goal:** Lift artifact-agnostic OCI helpers out of `oci/skills` into a shared `oci/artifact` subpackage so `oci/plugins` consumes them instead of copying. Behavior-preserving refactor — must preserve `oci/skills` public behavior exactly.

> **Implementation complete; PR incoming.** This issue exists for tracking and will be auto-closed by the PR that lands the refactor.

**Deliverables:**
- New exported package `toolhive-core/oci/artifact`.
- Move from `oci/skills`: `tar.go` (`CreateTar`, `TarOptions`, `FileEntry`, `DefaultTarOptions`), `gzip.go` (`Compress`/`DecompressWithLimit`), the extraction/`validatingTarget` pull-hardening + size/file-count caps, and (recommended) OCI platform helpers (`PlatformString`/`ParsePlatform`/`DefaultPlatforms`).
- Public-API preservation: `oci/skills` re-exports every moved symbol (type aliases, var-forwarding, const re-decl) so downstream `toolhive` imports are unaffected.
- `oci/skills` keeps `mediatypes.go` (skill-specific) but calls the shared package.
- **Out of scope:** the local-build marker (`dev.stacklok.toolhive.local-build`) lives in `toolhive`, not core — handled in Phase 2.

**Dependency/gate:** none upstream. Precedes **GATE-C1** (tag `toolhive-core`, e.g. `v0.0.25`; bump `toolhive/go.mod`).

**Testing/exit gate:** all pre-existing `oci/skills` unit + integration tests pass unchanged; determinism test still green (must not change published skill digests).

Part of stacklok/toolhive#5525
Refs RFC PR stacklok/toolhive-rfcs#77

JAORMX:
## Phase 0 of plugin support (RFC THV-0077)

Extracts the artifact-agnostic OCI primitives out of `oci/skills` into a new `oci/artifact` package so the upcoming `oci/plugins` package can reuse them. **Behavior-preserving** refactor — no logic changes.

### What moved to `oci/artifact`
- **tar** — `CreateTar`/`ExtractTar`/`ExtractTarWithLimit`, `FileEntry`, `TarOptions`/`DefaultTarOptions`, `MaxTarFileSize`, path-traversal/symlink rejection
- **gzip** — `Compress`/`Decompress`/`DecompressWithLimit`, `CompressTar`/`DecompressTar`, `GzipOptions`/`DefaultGzipOptions`, `MaxDecompressedSize` (decompression-bomb guard)
- **platform** — `PlatformString`/`ParsePlatform`/`DefaultPlatforms` + OS/Arch consts
- **pull-hardening** — promoted to exported `ValidatingTarget`/`NewValidatingTarget` (+ manifest/blob size caps and manifest-count limits, digest verification)

### Backward compatibility
`oci/skills` re-exports every moved symbol via type aliases (`type X = artifact.X`), `var`-forwarding for funcs, and const re-declaration (`oci/skills/artifact_aliases.go`). **The public surface of `oci/skills` is unchanged** — no exported signature changed, and downstream consumers (toolhive) are unaffected. The skills determinism tests still assert byte-stable artifact digests, proving published skill digests don't change.

### Notes
- Skill-specific code (media types, labels, `SkillConfig`, `Packager`, `Registry`, `Store`) stays in `oci/skills`.
- New package is **Alpha** (per CLAUDE.md stability tracks).
- This phase precedes **GATE-C1** (tag `toolhive-core` + bump `toolhive`'s `go.mod`), which is tracked in the epic.

Closes #130
Part of stacklok/toolhive#5525
RFC: stacklok/toolhive-rfcs#77


oci/skills/artifact_aliases.go
```diff
@@ -0,0 +1,86 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package skills
+
+import (
+	artifact "github.com/stacklok/toolhive-core/oci/artifact"
+)
+
+// This file re-exports the artifact-agnostic OCI primitives that were extracted
```
jhrozek, line 10: **[MEDIUM] Is the back-compat alias layer worth it for an Alpha package?**

`oci/skills` is marked **Alpha** in CLAUDE.md ("breaking changes possible"), so source compatibility isn't owed to callers yet. This layer (13 func forwarders + 3 type aliases + 7 const re-exports, plus the test-only mirror file) exists purely to avoid updating a handful of first-party importers (toolhive, dockyard, registries) — and it is the root cause of the duplicated-test/mirror-constant issue flagged in `artifact_aliases_test.go`.

Consider dropping the layer and migrating importers to `oci/artifact` directly. If a transition window is genuinely needed, keep it but mark the forwarders `// Deprecated:` so `go doc`/linters nudge migration — and still delete the duplicated skills-side tests.

oci/skills/artifact_aliases_test.go
```diff
@@ -0,0 +1,31 @@
...
+// Test-only forwarders for unexported artifact primitives that the existing
+// oci/skills test suite (registry_test.go, gzip_test.go) references directly.
+// Keeping these as test aliases lets the regression-gate tests stay unchanged
+// after the primitives were moved into the oci/artifact package.
+
+// gzipOSUnknown mirrors the gzip "unknown" OS header value used in tests.
+const gzipOSUnknown = 255
+
+// maxIndexManifests mirrors the artifact cap on manifests in an image index.
+const maxIndexManifests = 32
```
jhrozek: **[MEDIUM] Duplicated test suites + drifting hardcoded mirror constants**

These constants (`maxIndexManifests = 32`, `maxManifestLayers = 64`, `gzipOSUnknown = 255`) are hardcoded copies of values that are unexported in `oci/artifact`. There is no compiler link between them. If the real cap in `oci/artifact` changes (e.g. 32 -> 16), the skills-side boundary tests at `registry_test.go:176,191` (`maxIndexManifests+1`) keep testing the stale value and can **silently false-pass**.

Root cause: the tar/gzip/validation tests were copied verbatim into both `oci/artifact` and `oci/skills`; the skills copies only assert "the forwarder forwards." Recommend deleting the moved-logic tests from `oci/skills` (`gzip_test.go`, `tar_test.go`, and the manifest-count/validating-target cases in `registry_test.go`) — `oci/artifact` now owns them — which removes the need for these mirror constants entirely.

oci/artifact/validate.go
```diff
@@ -0,0 +1,155 @@
...
+	"oras.land/oras-go/v2"
+)
+
+// MaxManifestSize is the maximum size of a manifest (1MB).
+const MaxManifestSize int64 = 1 * 1024 * 1024
+
+// MaxBlobSize is the maximum size of a blob (100MB).
+const MaxBlobSize int64 = 100 * 1024 * 1024
+
+// maxIndexManifests is the maximum number of manifests in an image index.
```
jhrozek, line 24: **[LOW] Half-open export surface**

`ValidatingTarget`, `ValidateManifestCounts`, `IsManifestMediaType`, `MaxManifestSize`, and `MaxBlobSize` are exported, but the caps actually enforced here (`maxIndexManifests`, `maxManifestLayers`) stay unexported — so a caller of the exported `ValidateManifestCounts` can't introspect or configure the limits it applies. The only entry point `registry.go` actually uses is `NewValidatingTarget`.

Suggest deciding intentionally: either (a) unexport `ValidateManifestCounts`/`IsManifestMediaType` (shrinking the surface to `NewValidatingTarget`), or (b) commit to a public validation API and also export `MaxIndexManifests`/`MaxManifestLayers`. (a) is more honest to the single real use site — the two functions appear exported mainly to satisfy the moved skills tests.

jhrozek (commented):
Batched review from `/review-iterate` on the oci/artifact extraction. These are non-blocking (the refactor itself is sound — build/vet/test/lint/license all pass and the security posture is preserved). Findings cluster on the back-compat alias layer for an Alpha package, plus two pre-existing path-handling issues carried over by the tar.go rename. Test-coverage gaps from the review were noted but not commented inline.

oci/artifact/tar.go
```diff
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-package skills
+package artifact
```
jhrozek, line 4: Two **pre-existing** path-handling issues in this renamed file (carried over verbatim from `oci/skills`, not introduced here — flagging since the move is a reasonable moment to fix). The relevant code isn't in the rename's diff hunk, hence anchored here.

**[MEDIUM] `validateTarPath` (line ~161) uses `path`, not `path/filepath` — Windows `\` bypass.** `path.Clean` only treats `/` as a separator, so `path.Clean("foo\\..\\..\\etc\\passwd")` is returned unchanged and slips past the `HasPrefix(cleaned, "..")` check on a Windows host. The same prefix check also false-rejects a legit file named `..hidden`. Fix: reject any `\` (OCI/POSIX mandate `/`) and tighten to `cleaned == ".." || strings.HasPrefix(cleaned, "../")`.

**[LOW] `ExtractTar` (line ~151) stores raw `hdr.Name`, not the cleaned path.** Validation runs first, but the uncleaned name is stored in `FileEntry.Path`, so a caller doing string concatenation (rather than `filepath.Join`) could re-introduce surprising paths from e.g. `foo/./bar`. Fix: store `path.Clean(hdr.Name)`.

Severity is MEDIUM/LOW because this library targets Linux/k8s and `ExtractTar` returns in-memory entries (no FS write).

jhrozek (approved):
All 5 commented findings addressed. LGTM.

PR #136, opened 2026-06-18
oci/plugins: Phase 1 OCI artifact package for plugins (THV-0077)

### LINKED ISSUE #131: [plugins] Phase 1: oci/plugins package (THV-0077)
**Goal:** Add an `oci/plugins` package mirroring `oci/skills` file-for-file, built on the shared `oci/artifact` primitives from Phase 0.

**Deliverables:**
- `oci/plugins/mediatypes.go` — `ArtifactTypePlugin = "dev.toolhive.plugins.v1"`; `dev.toolhive.plugins.*` labels/annotations (`.name`, `.description`, `.version`, `.license`, `.files`, `.components`, `.requires`); `PluginConfig` + `PluginConfigFromImageConfig`. `.components` JSON inventory and `.requires` are plugin-specific additions.
- `oci/plugins/interfaces.go` — `RegistryClient` (Push/Pull), `PluginPackager`.
- `oci/plugins/packager.go` — index → per-platform manifest → config + single `plugin.tar.gz` layer, using `oci/artifact`.
- `oci/plugins/registry.go` — ORAS push/pull + Docker credential auth.
- `oci/plugins/store.go` — local store rooted at `toolhive/plugins` under `xdg.DataHome`.
- `oci/plugins/mocks/` via `go:generate mockgen`.

**Dependency/gate:** depends on **Phase 0** merged. Precedes **GATE-C2** (tag `toolhive-core`, e.g. `v0.0.26`; bump `toolhive/go.mod`). Nothing in Phase 2 compiles before C2.

**Testing/exit gate:** packager determinism (identical bundle → identical digest under `SOURCE_DATE_EPOCH`); config round-trip; store put/get.

Part of stacklok/toolhive#5525
Refs RFC PR stacklok/toolhive-rfcs#77

JAORMX:
## What

Adds the **`oci/plugins`** package — Phase 1 of the plugin lifecycle epic (THV-0077). It mirrors `oci/skills` file-for-file, built on the shared `oci/artifact` primitives extracted in Phase 0.

A *plugin* is a directory declared by `.claude-plugin/plugin.json`, bundling slash commands, subagents, Agent Skills, hooks, and MCP/LSP server configs. This package builds such a directory into a reproducible, content-addressable multi-platform OCI artifact, pushes/pulls it via ORAS, and stores it locally.

## Deliverables (per stacklok/toolhive-core#131)

- **`mediatypes.go`** — `ArtifactTypePlugin = "dev.toolhive.plugins.v1"`; full `dev.toolhive.plugins.*` label/annotation set (`.name`, `.description`, `.version`, `.license`, `.files`, plus plugin-specific `.components` inventory and `.requires`); `PluginConfig` + `PluginConfigFromImageConfig`.
- **`interfaces.go`** — `RegistryClient` (Push/Pull), `PluginPackager`.
- **`packager.go`** — index → per-platform manifest → config + single `plugin.tar.gz` layer; `SOURCE_DATE_EPOCH`-aware for reproducible digests.
- **`registry.go`** — ORAS push/pull + Docker credential auth.
- **`store.go`** — local OCI store rooted at `toolhive/plugins` under `xdg.DataHome`.
- **`mocks/`** — generated via `go:generate mockgen`.

## Testing

Exit gates covered: packager determinism under `SOURCE_DATE_EPOCH`, config round-trip (incl. zero-component nil regression guard), store put/get. Plus end-to-end `package → push → pull → extract` integration tests mirroring `oci/skills`.

- `task lint` → 0 issues
- `task test` (race) → all pass
- coverage **73.6%** (>70% graduation bar)
- `task license-check` → clean

## Review

A multi-axis review panel (spec / standards / architecture / security / devex / duplication) ran on the diff. No ship-blockers. In-scope findings applied: `componentInventory` now returns `nil` when empty (round-trip fix), a dedicated config round-trip test, and `integration_test.go`. Deferred (parity with `oci/skills`, separate follow-ups): extracting the byte-identical `Store`/`Registry` into `oci/artifact`, and `Pull` SSRF/`PlainHTTP` hardening.

## Next gate

Precedes **GATE-C2**: tag `toolhive-core` after merge, then bump `toolhive/go.mod`. Nothing in Phase 2 (stacklok/toolhive#5526) compiles before that tag.

Part of stacklok/toolhive#5525
Closes stacklok/toolhive-core#131


oci/plugins/store.go
```diff
@@ -0,0 +1,338 @@
...
+			return err
+		}
+	}
+	return s.deleteBlob(d)
+}
+
+// deleteBlob removes the blob file for d from the local OCI layout.
+// A missing file is treated as success (idempotent).
+func (s *Store) deleteBlob(d digest.Digest) error {
+	path := filepath.Join(s.root, "blobs", d.Algorithm().String(), d.Encoded())
```
jhrozek: **[BLOCKER B1] Path traversal via unvalidated digest — arbitrary file deletion**

`d.Algorithm().String()` and `d.Encoded()` are taken from user-supplied JSON (`m.Config.Digest`, `m.Layers[i].Digest` parsed in `deleteManifestBlobs`) without calling `d.Validate()`. `digest.Digest` is an unvalidated `string` typedef. A crafted digest such as `"sha256:../../../../../etc/passwd"` causes `filepath.Join` to produce a path escaping the store root, and `os.Remove` deletes an arbitrary file the process can reach. The oras read/write path validates digests internally — only this custom `os.Remove` path is exposed.

Same defect in `oci/skills/store.go`.

```suggestion
func (s *Store) deleteBlob(d digest.Digest) error {
	if err := d.Validate(); err != nil {
		return fmt.Errorf("deleting blob: invalid digest %q: %w", d, err)
	}
	blobRoot := filepath.Join(s.root, "blobs")
	path := filepath.Join(blobRoot, d.Algorithm().String(), d.Encoded())
	if !strings.HasPrefix(filepath.Clean(path)+string(filepath.Separator),
		filepath.Clean(blobRoot)+string(filepath.Separator)) {
		return fmt.Errorf("deleting blob: path escapes store root: %s", path)
	}
	if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
		return fmt.Errorf("deleting blob %s: %w", d, err)
	}
	return nil
}
```

Note: add `"strings"` to the import block.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+}
+
+// readPluginDirectory reads a plugin directory, validates its contents, and parses the manifest.
+func readPluginDirectory(dir string) (*pluginDirContent, error) {
+	if err := validatePluginDir(dir); err != nil {
+		return nil, err
+	}
+
+	manifestPath := filepath.Join(dir, ManifestFileName)
+	manifest, err := os.ReadFile(manifestPath) //#nosec G304 -- path constructed from user-provided plugin directory
```
jhrozek: **[BLOCKER B2] `os.ReadFile` before size check + missing `Lstat` guard**

Two issues at this line:

1. **Memory exhaustion**: `os.ReadFile` reads the entire file into memory unconditionally. The `maxManifestSize` check lives inside `parseManifest` (called after the read). A 1 GB `plugin.json` is fully allocated before rejection.

2. **Symlink TOCTOU**: `collectPluginFile` uses `os.Lstat` to reject symlinks on walked files, but the manifest is read here — before the walk — with no `Lstat` guard. An attacker with write access can race a symlink replacement of `.claude-plugin/plugin.json` between `validatePluginDir` and this `ReadFile`, embedding arbitrary file content in the OCI layer.

Fix: insert `os.Lstat` + size check before the read:

```suggestion
	fi, err := os.Lstat(manifestPath)
	if err != nil {
		if os.IsNotExist(err) {
			return nil, fmt.Errorf("%s not found in plugin directory: %w", ManifestFileName, ErrPluginManifestMissing)
		}
		return nil, fmt.Errorf("checking manifest: %w", err)
	}
	if fi.Mode()&os.ModeSymlink != 0 || !fi.Mode().IsRegular() {
		return nil, fmt.Errorf("manifest must be a regular file: %w", ErrInvalidPluginFile)
	}
	if fi.Size() > maxManifestSize {
		return nil, fmt.Errorf("manifest file size %d exceeds maximum of %d bytes: %w",
			fi.Size(), maxManifestSize, ErrInvalidPluginManifest)
	}
	manifest, err := os.ReadFile(manifestPath) //#nosec G304 -- path within validated plugin directory; Lstat guard above prevents symlink follow
```

oci/plugins/registry.go
```diff
@@ -0,0 +1,167 @@
...
+	target, err := r.newTarget(parsedRef)
+	if err != nil {
+		return "", fmt.Errorf("getting repository: %w", err)
+	}
+
+	validated := artifact.NewValidatingTarget(store.Target())
+
+	// Copy from remote to the validated local store.
+	desc, err := oras.Copy(
+		ctx, target, parsedRef.Reference, validated, parsedRef.Reference, oras.DefaultCopyOptions,
```
jhrozek: **[BLOCKER B3] `Pull` stores the short ref as the local tag — cross-plugin tag collision**

`oras.Copy` is called with `parsedRef.Reference` (the bare tag, e.g. `v1.0.0`) as the destination tag. The local store at `~/.local/share/toolhive/plugins` is shared across all plugins. Pulling `example.com/orgA/plugin:v1.0.0` then `example.com/orgB/plugin:v1.0.0` silently overwrites the `v1.0.0` tag to point at the second plugin's digest. The `store.Tag(ctx, desc.Digest, ref)` on line 133 adds the full ref afterwards, but the poisoned short-ref tag remains.

Fix: pass the full `ref` as the destination tag so `oras.Copy` stores it under the canonical full reference, then drop the separate `store.Tag` call:

```suggestion
		ctx, target, parsedRef.Reference, validated, ref, oras.DefaultCopyOptions,
```

Then remove lines 131–135. Same bug in `oci/skills/registry.go`.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+		if os.IsNotExist(err) {
+			return fmt.Errorf("plugin directory not found: %s: %w", dir, ErrInvalidPluginDir)
+		}
+		return fmt.Errorf("accessing plugin directory: %w: %w", err, ErrInvalidPluginDir)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("path is not a directory: %s: %w", dir, ErrInvalidPluginDir)
+	}
+
+	cleanDir := filepath.Clean(dir)
```
jhrozek, line 274: **[MEDIUM] Path-traversal check is logically ineffective after `filepath.Clean`**

`filepath.Clean` resolves all `..` segments before the check runs. `filepath.Clean("/tmp/../../etc/passwd")` returns `"/etc/passwd"` — the `strings.Contains` check never fires for a traversal that resolves to a valid absolute path. The only case it catches is an unresolvable relative path, which `os.Stat` already rejects.

Replace with an absolute-path guard:

```suggestion
	cleanDir := filepath.Clean(dir)
	if !filepath.IsAbs(cleanDir) {
		return fmt.Errorf("plugin directory must be an absolute path: %w", ErrInvalidPluginDir)
	}
```

oci/plugins/store.go
```diff
@@ -0,0 +1,338 @@
...
+// fetchContent retrieves raw content by digest from the underlying store.
+func (s *Store) fetchContent(ctx context.Context, d digest.Digest) ([]byte, error) {
+	// oci.Store's Fetch only uses the Digest field to locate blobs in blobs/<algo>/<hex>.
+	rc, err := s.inner.Fetch(ctx, ocispec.Descriptor{Digest: d})
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = rc.Close() }()
+
+	data, err := io.ReadAll(rc)
```
jhrozek: **[MEDIUM] `fetchContent` uses unbounded `io.ReadAll` on local blobs**

The descriptor passed to `inner.Fetch` carries only a `Digest` (no `Size`), and the reader is consumed with no size cap. If the local OCI layout is corrupted or replaced, any call to `GetManifest`, `GetIndex`, or `IsIndex` attempts an unbounded allocation. The `ValidatingTarget` wrapper guards against large *incoming* payloads on pull — local reads need the same protection.

```suggestion
	data, err := io.ReadAll(io.LimitReader(rc, artifact.MaxBlobSize+1))
	if err != nil {
		return nil, err
	}
	if int64(len(data)) > artifact.MaxBlobSize {
		return nil, fmt.Errorf("blob %s exceeds local fetch size limit", d)
	}
```

`artifact` is already imported. Same fix in `oci/skills/store.go`.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+
+	for _, p := range sortedPaths {
+		files = append(files, artifact.FileEntry{
+			Path:    p,
+			Content: content.files[p],
+		})
+	}
+
+	tarOpts := artifact.TarOptions{Epoch: opts.Epoch}
+	gzipOpts := artifact.DefaultGzipOptions()
```
jhrozek, line 422: **[MEDIUM] `SOURCE_DATE_EPOCH` is not applied to the gzip member header**

`artifact.DefaultGzipOptions()` hard-codes `Epoch: time.Unix(0, 0).UTC()`. Tar headers and OCI annotations/config honour `opts.Epoch`, but the gzip `ModTime` is always 1970-01-01 regardless of `SOURCE_DATE_EPOCH`. The digest is still deterministic so reproducibility tests pass — but a consumer inspecting the gzip header sees a timestamp disagreeing with every other timestamp in the artifact.

```suggestion
	gzipOpts := artifact.DefaultGzipOptions()
	gzipOpts.Epoch = opts.Epoch
```

Same fix in `oci/skills/packager.go`.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+
+	return PackageOptions{
+		Epoch:     epoch,
+		Platforms: artifact.DefaultPlatforms,
+	}
+}
+
+// Package packages a plugin directory into an OCI artifact in the local store.
+func (p *Packager) Package(ctx context.Context, pluginDir string, opts PackageOptions) (*PackageResult, error) {
+	if len(opts.Platforms) == 0 {
```
jhrozek, line 120: **[MEDIUM] Zero-value `PackageOptions.Epoch` produces inconsistent timestamps**

With `PackageOptions{}` the timestamp sources disagree:
- `artifact.CreateTar` normalises a zero epoch to `time.Unix(0,0).UTC()` → tar mtime = 1970
- `artifact.Compress` similarly defaults to 1970
- `createOCIConfig` / `createManifest` use `opts.Epoch` as-is → `Image.Created` marshals to `"0001-01-01T00:00:00Z"`

Result: config says year 1, layer says 1970. `DefaultPackageOptions()` avoids this but `PackageOptions{}` is a valid zero value.

```suggestion
	if len(opts.Platforms) == 0 {
		opts.Platforms = artifact.DefaultPlatforms
	}
	if opts.Epoch.IsZero() {
		opts.Epoch = time.Unix(0, 0).UTC()
	}
```

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+	// maxManifestSize limits plugin.json to prevent JSON parsing attacks.
+	maxManifestSize = 64 * 1024
+
+	// maxPluginFiles limits the number of files in a plugin directory to prevent
+	// memory exhaustion during packaging.
+	maxPluginFiles = 1_000
+
+	// maxPluginTotalSize limits the total aggregate size of all files in a plugin
+	// directory to prevent memory exhaustion during packaging (100 MB).
+	maxPluginTotalSize int64 = 100 * 1024 * 1024
```
jhrozek: **[MEDIUM] `maxPluginTotalSize` + tar overhead can exceed `MaxDecompressedSize` — artifacts you package cannot be unpacked**

`collectPluginFiles` admits up to 100 MiB of file content. `artifact.CreateTar` adds ~512 bytes of header per file; with `maxPluginFiles = 1_000` that is ~0.5 MiB of overhead, pushing the uncompressed tar above `artifact.MaxDecompressedSize` (100 MiB). The extraction path enforces that limit via `DecompressWithLimit`. A near-full plugin packages successfully but fails to unpack.

The two limits must be co-ordinated. Simplest fix:

```suggestion
	maxPluginTotalSize int64 = 95 * 1024 * 1024
```

Same issue in `oci/skills`.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+
+	return &pm, nil
+}
+
+// createContentLayer creates a reproducible tar.gz of the plugin content.
+// Returns both compressed and uncompressed bytes (uncompressed needed for diff_id).
+func createContentLayer(content *pluginDirContent, opts PackageOptions) (compressed, uncompressed []byte, err error) {
+	var files []artifact.FileEntry
+
+	files = append(files, artifact.FileEntry{
```
jhrozek, line 401: **[MEDIUM] File modes stripped to 0644 — executable scripts lose their mode**

`artifact.FileEntry` has a `Mode` field, but all entries are built with only `Path` and `Content`. `artifact.CreateTar` defaults every entry to `0644`. A plugin's `scripts/hook.sh` written as `0755` is silently flattened and will not be executable after extraction. The test fixture includes a `#!/bin/sh` script that exercises this path.

Fix: capture `fileInfo.Mode()` in `collectPluginFile`, store it alongside the content, and populate `Mode` when constructing `FileEntry` entries here and in the loop at line 373.

oci/plugins/registry.go
```diff
@@ -0,0 +1,167 @@
...
+// defaultNewTarget creates a remote repository client for the given parsed reference.
+func (r *Registry) defaultNewTarget(ref registry.Reference) (oras.Target, error) {
+	repoPath := ref.Registry + "/" + ref.Repository
+
+	repo, err := remote.NewRepository(repoPath)
+	if err != nil {
+		return nil, fmt.Errorf("creating repository for %q: %w", repoPath, err)
+	}
+
+	repo.Client = &auth.Client{
```
jhrozek, line 159: **[MEDIUM] No HTTP retry client — transient registry errors fail permanently**

`auth.Client{Credential: ...}` leaves the inner `Client` as `nil`, resolving to `http.DefaultClient` with no retries. Transient 5xx, connection resets, or 429s from ghcr.io/ECR under load fail push/pull immediately. The ORAS-documented pattern is `retry.DefaultClient`.

```suggestion
	repo.Client = &auth.Client{
		Client:     retry.DefaultClient,
		Credential: credentials.Credential(r.credStore),
	}
```

Requires adding `"oras.land/oras-go/v2/registry/remote/retry"` to imports. Same fix in `oci/skills/registry.go`.

oci/plugins/store.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+		return fmt.Errorf("removing tag: %w", err)
+	}
+	return nil
+}
+
+// DeleteBuild removes a tag and, if no other tag shares the same digest,
+// deletes all associated blobs (config, layers, manifest, and index if applicable).
+// Use DeleteTag when tag-only removal is desired and blob cleanup is not needed.
+func (s *Store) DeleteBuild(ctx context.Context, tag string) error {
```
jhrozek, line 176: **[MEDIUM] `DeleteBuild` read-modify-write is not atomic**

The sequence `Resolve → DeleteTag → isDigestReferenced → deleteOrphanedBlobs` is not protected by a mutex. A `DeleteBuild` racing a concurrent `Tag` on the same digest can delete blobs that the newly-added tag now references. `deleteBlob` is idempotent so double-invocation is safe, but the race between tag addition and blob deletion is not.

Add a `sync.Mutex` to `Store` and hold a write lock across `DeleteBuild`, `DeleteTag`, and `Tag`.

oci/plugins/packager.go
```diff
@@ -0,0 +1,624 @@
...
+// createOCIConfig creates the OCI image config with plugin metadata in labels.
+func createOCIConfig(
+	content *pluginDirContent,
+	uncompressedTar []byte,
+	platform ocispec.Platform,
+	opts PackageOptions,
+) (*ocispec.Image, *PluginConfig) {
+	cfg := pluginConfig(content)
+
+	filesJSON, _ := json.Marshal(cfg.Files)
```
jhrozek: **[LOW] `json.Marshal` errors silently discarded**

The blank `_` is safe today because `[]string` and `map[string]int` cannot produce marshal errors. But a future refactor widening `ComponentInventory` to an interface type would silently produce broken OCI labels. Same pattern in `createManifest` at line 539.

Use an explicit panic for truly invariant marshalling:

```suggestion
	filesJSON, err := json.Marshal(cfg.Files)
	if err != nil {
		panic(fmt.Sprintf("plugins: marshal files: %v", err))
	}
	componentsJSON, err := json.Marshal(cfg.Components)
	if err != nil {
		panic(fmt.Sprintf("plugins: marshal components: %v", err))
	}
	requiresJSON, err := json.Marshal(cfg.Requires)
	if err != nil {
		panic(fmt.Sprintf("plugins: marshal requires: %v", err))
	}
```

oci/plugins/store_test.go
```diff
@@ -0,0 +1,315 @@
...
+func TestStore_DeleteTag_NotFound(t *testing.T) {
+	t.Parallel()
+
+	store, err := NewStore(t.TempDir())
+	require.NoError(t, err)
+
+	err = store.DeleteTag(context.Background(), "missing")
+	require.Error(t, err)
+	assert.Equal(t, http.StatusNotFound, httperr.Code(err))
+}
```
jhrozek, line 302: **[MEDIUM] `DeleteBuild` and all blob-deletion helpers are 0% covered**

`DeleteBuild`, `isDigestReferenced`, `deleteOrphanedBlobs`, `deleteManifestBlobs`, and `deleteBlob` all report 0.0% in `go tool cover -func`. This is the non-trivial orphan-graph-walk logic that most needs regression tests.

`oci/skills/store_test.go` has a complete `TestStore_DeleteBuild` covering:
1. Tag removed + all blobs deleted when no other tag shares the digest
2. Blobs retained when a second tag aliases the same digest
3. 404 on nonexistent tag

Please port `putTestArtifact`, `blobExists`, and `TestStore_DeleteBuild` from `oci/skills/store_test.go`, and add a 4th subcase using an `ocispec.Index` to cover the `deleteOrphanedBlobs` index branch — that path is untested in both packages.

jhrozek (changes requested):
## Review: `oci/plugins` package

This review combines findings from three parallel specialist agents (security, Go quality, architecture) plus the MoE review in `REVIEW-e36cae86.md`. The package is a well-structured mirror of `oci/skills` and the plugin-specific logic is clean. Three issues should be addressed before merge.

### Blockers
- **B1** — path traversal in `deleteBlob` via unvalidated digest (arbitrary file deletion)
- **B2** — `os.ReadFile` before size check enables memory exhaustion; same line lacks `Lstat` guard (symlink TOCTOU)
- **B3** — `Pull` uses short ref as local tag, causing cross-plugin tag collisions in the shared store

### Other significant findings
- `validatePluginDir` path-traversal check is logically ineffective after `filepath.Clean`
- `fetchContent` uses unbounded `io.ReadAll` on local blobs
- `SOURCE_DATE_EPOCH` not applied to gzip member header
- Zero-value `PackageOptions.Epoch` produces year-1 timestamps in config but 1970 in tar/gzip
- `maxPluginTotalSize` (100 MB) + tar overhead can exceed `MaxDecompressedSize` (100 MB)
- File modes stripped to 0644 — executable scripts lose their mode
- No HTTP retry client in `defaultNewTarget`
- `DeleteBuild` read-modify-write is not atomic
- `json.Marshal` errors silently discarded in config/manifest creation
- `TestStore_DeleteBuild` entirely absent — blob-deletion surface (5 functions) is 0% covered

Findings marked "same in skills" should be fixed in both packages.

JAORMX:
Thanks for the thorough review @jhrozek. All findings addressed in 169911d. Summary:

**Blockers (plugins)**
- **B1** — `Store.deleteBlob` now calls `d.Validate()` and guards the resolved path against escaping the store root before `os.Remove`.
- **B2** — `readPluginDirectory` now `Lstat`s the manifest, rejects a symlink/non-regular file, and enforces `maxManifestSize` before `os.ReadFile`.
- **B3** — `Pull` passes the full `ref` as the `oras.Copy` destination tag and drops the separate `store.Tag`, so a bare tag can no longer collide across artifacts in the shared store.

**Other findings (plugins)**
- `validatePluginDir` — replaced the inert post-`Clean` `..` check with an absolute-path guard.
- `fetchContent` — local reads bounded with `io.LimitReader(artifact.MaxBlobSize+1)`.
- gzip member header now honours `opts.Epoch`.
- zero-value `PackageOptions.Epoch` normalized to Unix 0 so config/tar/gzip timestamps agree.
- `maxPluginTotalSize` → 95 MB, below `artifact.MaxDecompressedSize`, so tar overhead can't make a packaged artifact un-unpackable.
- file modes preserved through packaging (executable scripts keep their bit); added a regression test.
- `retry.DefaultClient` wired into the auth client for transient registry errors.
- `DeleteBuild`/`DeleteTag`/`Tag` serialized under a `sync.Mutex` (non-reentrant: public methods delegate to `*Locked` helpers).
- `json.Marshal` of labels/annotations now panics on the invariant error via a `mustMarshalJSON` helper.

**Tests** — ported `TestStore_DeleteBuild`, added an `ocispec.Index` subcase covering the `deleteOrphanedBlobs` index branch, a `deleteBlob` invalid-digest guard, and a file-mode preservation test. Plugins coverage 73.6% → **81.7%**.

**Parity (oci/skills)** — applied the findings you marked "same in skills": B1, B3, `fetchContent` bound, gzip epoch, total-size limit, and the retry client.

A couple of skills-only parity items I did **not** pull into this PR, to keep it scoped — happy to do either here or as a follow-up if you'd prefer:
- B2's `Lstat`/size guard on `SKILL.md` (skills has its own frontmatter path; same TOCTOU class applies).
- The `DeleteBuild` mutex (the race exists in skills too).

`task` is green (lint 0 issues, vet, race tests, license headers).

jhrozek (approved):

PR #157, opened 2026-07-08
fix(mcpcompat): close functional gaps vs mcp-go for loss-free migration (#156)

### LINKED ISSUE #156: Close mcpcompat functional gaps vs mcp-go for a loss-free migration
## Context

stacklok/toolhive#5729 migrates ToolHive from `mark3labs/mcp-go v0.55.1` onto the `mcpcompat` shim (introduced in #150, released in v0.0.27, backed by `modelcontextprotocol/go-sdk v1.6.1`). A code-level audit of the shim against ToolHive's actual mcp-go usage found functional regressions — some disclosed in that PR, some not. ToolHive wants to complete the migration (it is the path to MCP 2026-07-28 / stateless support via go-sdk v1.7.x) **without accepting any functional loss**.

We verified that **no newer go-sdk release helps**: latest stable is v1.6.1, latest overall is v1.7.0-pre.1, and the relevant API surface is identical in both. Every fix below is therefore shim-side work against v1.6.1, except where noted as an upstream ask.

## Audit findings (shim v0.0.27, file references are `mcpcompat/`)

| # | Finding | Status | Evidence |
|---|---------|--------|----------|
| 1a | Client never registers `ProgressNotificationHandler` / `LoggingMessageHandler` (both exist in go-sdk v1.6.1 `ClientOptions`) — only the 3 `list_changed` handlers are wired. Combined with 1b, **no server-initiated notification survives the ToolHive stdio bridge end-to-end** | **Lost — shim bug** | `client.go:417-427` |
| 1b | Server side forwards only progress + message notifications; `tools/list_changed` and everything else is dropped (go-sdk has no public generic notification sender) | **Lost — upstream API gap** | `notifications.go:32-76` |
| 2 | `WithHeartbeatInterval` is a silent no-op: the interval is stored and never used; go-sdk `StreamableHTTPOptions` has no keep-alive. Long-lived SSE streams get no pings and can be dropped by LB idle timeouts | **Lost — undisclosed** | `transports.go:85,142-143` |
| 3 | `SessionIdManager` is largely wired (Generate via `GetSessionID`, Terminate via DELETE, Validate on cross-replica rehydration) but per-request `Validate` is **skipped for local sessions** — mcp-go validated every request. A session terminated cross-pod or via auth-failure can keep being served by the origin pod | **Degraded** | `server.go:334-336`, `transports.go:154,199-208,216,233` |
| 4 | 401/auth errors are string-matched (`"401"`, `"unauthorized"`); `extractResourceMetadataURL` always returns `""`; 403-on-initialize is not mapped to `ErrLegacySSEServer`. The sentinels ToolHive consumes are produced, so impact is low today | **Degraded** | `errors.go:66-93` |
| 5 | go-sdk enables DNS-rebinding/localhost protection by default; the shim does not expose `DisableLocalhostProtection`. Local proxies with custom Host headers now get 403 (mcp-go had no such protection) | **Behavior change** | `transports.go:166,170` |
| 6 | go-sdk paginates `tools/list` at `DefaultPageSize=1000`; mcp-go returned everything. Non-cursor-following clients see silent truncation past 1000 tools | **Behavior change** | go-sdk `server.go:35-36` |
| 7 | Per-session dynamic tool projection (security-relevant for vMCP): genuinely implemented — per-session go-sdk `Server` + live `AddTool`/`RemoveTools` reconciliation | **Kept** (needs runtime validation) | `server.go:298,375-423`, `session.go:122-239` |
| 8 | Doc comments claim progress/logging notifications are forwarded (they are not, see 1a) and the `WithSessionIdManager` comment says it is unwired (it mostly is wired) | **Docs wrong** | `client.go`, `notifications.go`, `transports.go:135-136` |

### Shim fixes (target: v0.0.28)

- [ ] **1a. Register client notification handlers** — wire go-sdk's `ProgressNotificationHandler` + `LoggingMessageHandler` into the mcpcompat client and route to the compat notification callback. Cheapest fix, largest correctness gain.
- [ ] **1b. `list_changed` via feature re-sync** — instead of forwarding the raw notification (impossible without a generic sender), re-sync the mirrored tools/resources/prompts on upstream `list_changed`; go-sdk auto-emits `list_changed` to clients on feature mutation. This restores (and arguably improves) mcp-go semantics without waiting on upstream.
- [ ] **2. Implement keep-alive shim-side** — honor `WithHeartbeatInterval` by injecting SSE comment lines (`: keepalive`) on a ticker into `text/event-stream` responses via a `ResponseWriter` wrapper around the go-sdk handler. SSE comments are spec-legal and ignored by clients.
- [ ] **3. Per-request `Validate` for local sessions** — `Mcp-Session-Id` is visible at the shim's HTTP layer before delegation to the go-sdk handler; call `Validate` on every request regardless of session locality, restoring mcp-go semantics (cross-pod/auth-failure termination takes effect immediately).
- [ ] **4. Adopt `StreamableClientTransport.OAuthHandler`** — replace 401 string-matching with go-sdk's native OAuth flow hook (exists in v1.6.1); populate `AuthorizationRequiredError.ResourceMetadataURL`; map 403-on-initialize to `ErrLegacySSEServer`.
- [ ] **5. Expose `DisableLocalhostProtection`** — plumb the option through shim config so callers can preserve pre-migration behavior for local proxies (default should keep protection on; ToolHive opts out only where it fronts localhost with custom Host headers).
- [ ] **6. Configurable server page size** — expose go-sdk's page size so aggregating servers (vMCP) can raise it; document that clients must follow cursors (client-side cursor loops are tracked in ToolHive, not here).
- [ ] **7. Fix the false doc comments** from finding 8.

### Upstream asks (file against `modelcontextprotocol/go-sdk`, non-blocking)

- [ ] Public generic notification sender on `ServerSession` (unblocks true notification passthrough; only typed `NotifyProgress`/`Log` exist today).
- [ ] Server-side keep-alive option in `StreamableHTTPOptions`.
- [ ] Typed HTTP-status errors from transports (removes string-matching).

### Acceptance criteria

- ToolHive adds regression tests (tracked in stacklok/toolhive) that fail against v0.0.27 and pass against v0.0.28: bridge end-to-end forwarding of progress/message/list_changed, cross-pod session termination taking effect on the origin pod, SSE keep-alive presence on idle streams, >1000-tool listing behavior.
- stacklok/toolhive#5729 bumps to the fixed release and merges with those tests green — i.e. the mcp-go → go-sdk migration lands with **zero functional loss**.
- Runtime validation of per-session tool projection (finding 7) performed as part of the ToolHive-side test work.

## Why now

The 2026-07-28 stateless MCP revision ships this month; go-sdk v1.7.x is the only Go SDK line with support for it. The shim is the bridge that lets ToolHive get there — these gaps are the only thing standing between us and a loss-free migration.

JAORMX:
## Summary

Closes #156 — closes the mcpcompat functional gaps vs mcp-go so the ToolHive migration (stacklok/toolhive#5729) lands with **zero functional loss**.

Five commits, one per wave, each reviewed by an independent panel (spec + standards + domain specialists + QA) with findings iterated before commit.

| Wave | Commit | Items |
|------|--------|-------|
| 1 | `b22b8e1` | **U1** (merge blocker) capability flags → `ServerOptions.Capabilities`; **U3** (merge blocker, security) per-request context bridge rekeyed to crypto/rand nonce; **U6** normalize only nil/empty input schemas + recover per-session AddTool panics |
| 2 | `a302089` | **1a+U8** register progress/logging notification handlers + carry params; **1b** confirm list_changed auto-emitted by go-sdk on feature mutation |
| 3 | `3c7ab69` | **2** `WithHeartbeatInterval` → `ServerOptions.KeepAlive`; **5** `WithDisableLocalhostProtection(bool)`; **6** `WithPageSize(n int)`; **3/U5** per-request `Validate` for local sessions |
| 4 | `19c482b` | **4/U7** scope 401 string-matching to transport-level errors (no false-positive OAuth refresh on JSON-RPC tool errors); restore any-4xx-on-initialize → `ErrLegacySSEServer` |
| 5 | `884cfd3` | **U4** elicitation surface (client `ElicitationHandler` + server `RequestElicitation`); **U9** populate before-hook request objects (tool name + args + cursor); **U2** document preset session ID limitation (upstream-only); **7** fix remaining stale doc comments |

### Shim fixes (all 15 work items from the issue + 2 audit comments)

- ✅ **U1** — `WithToolCapabilities`/`WithResourceCapabilities`/`WithPromptCapabilities` mapped to go-sdk `ServerOptions.Capabilities`; vMCP advertises tools/resources/prompts at initialize even with zero pre-registered features.
- ✅ **U3** — `pendingReqCtx` rekeyed from session-ID to a per-POST `crypto/rand` nonce (`X-MCP-Req-Nonce`); eliminates the cross-request context-bleed / identity-attribution race. go-sdk does not propagate per-POST context into session middleware (verified), so the bridge is retained but keyed per-request.
- ✅ **U6** — `normalizeObjectSchema` narrows to nil/empty-only; `$ref`/`oneOf`/boolean pass through verbatim; per-session `AddTool` panics recovered and skipped.
- ✅ **1a + U8** — `ProgressNotificationHandler` + `LoggingMessageHandler` wired in the client; `dispatch` carries params via `toNotificationParams` (JSON round-trip, `_meta` split).
- ✅ **1b** — Confirmed go-sdk auto-emits `tools/list_changed` on `SetSessionTools` (no production code needed beyond U1); corrected stale doc comments.
- ✅ **2** — `WithHeartbeatInterval` wired to `ServerOptions.KeepAlive` (was a silent no-op).
- ✅ **5** — `WithDisableLocalhostProtection(bool)` exposed (default: protection ON).
- ✅ **6** — `WithPageSize(n int)` exposed (default 0 → go-sdk's 1000).
- ✅ **3/U5** — `SessionIdManager.Validate` called for local sessions too; restores per-request liveness + sliding TTL.
- ✅ **4/U7** — 401 string-matching scoped to transport-level errors via captured HTTP status + `errors.As(*jsonrpc.Error)`; false-positive OAuth refresh on tool-error text eliminated; any-4xx-on-initialize → `ErrLegacySSEServer` restored.
- ✅ **U4** — Elicitation surface added (client `ElicitationHandler`/`WithElicitationHandler`/`NewStreamableHttpClientWithOpts`; server `SessionWithElicitation`/`RequestElicitation`/`ErrNoActiveSession`/`ErrElicitationNotSupported`); JSONResponse mode documented (requires client continuous listening).
- ✅ **U9** — Before-hook request objects populated (`mcp.CallToolRequest` with name + args, `mcp.ListToolsRequest` with cursor); latent `translateUnknownToolError` type-assertion bug fixed (`*CallToolParams` → `*CallToolParamsRaw`).
- ✅ **U2** — Preset session ID limitation documented loudly (upstream-only: go-sdk has no `SessionID` field on `StreamableClientTransport`; honored via resume path only).
- ✅ **7** — All stale doc comments fixed (`WithSessionIdManager`, package doc, `SessionWithTools`, capability-flag comment, `OnNotification`, `notifications.go`, `normalizeObjectSchema`).

### Upstream asks (documented, not implemented — tracked in the issue)

- Public generic notification sender on `ServerSession`.
- Preset session ID on `StreamableClientTransport` (blocks U2).
- Per-request session `Validate`/`Terminate` hooks.
- Typed HTTP-status errors from transports.
- Per-request context propagation to stateful session handlers (U3 root cause).

## Test coverage

Each item has e2e and/or unit tests that would fail without the fix:
- Capability advertisement e2e (`TestAdvertisesToolsCapability_WithNoGlobalTools`)
- Per-request context isolation under `-race` (verified to fail without the bridge)
- Schema normalization table + non-object schema panic recovery
- Progress/logging notification e2e with param assertions
- `list_changed` e2e on `SetSessionTools`
- `toNotificationParams` unit table (meta split, error paths, nil guards)
- Heartbeat eviction e2e
- Localhost-protection toggle e2e (raw non-localhost Host)
- Page-size pagination e2e with `nextCursor`
- Local-session validate-evicts-on-shared-terminate
- False-positive JSON-RPC error isolation + transport 401/404/403 classification + 4xx-on-initialize legacy SSE
- Elicitation e2e (works with continuous listening, fails without) + handler error return + guard branches
- Before-hook tool name + args + cursor
- Unknown-tool error contains tool name (regression test for `translateUnknownToolError`)

## Verification

- `task lint` — 0 issues (golangci-lint + go vet)
- `task test` — all packages pass with `-race`
- `task license-check` — clean

## Review process

Each wave was independently reviewed by a panel (spec adherence, standards conformance, secure-code-reviewer, software-architect, devex-reviewer, QA test-expert) in fresh contexts. Cross-confirmed findings were iterated before commit:
- Wave 1: U6 per-session panic recovery (cross-confirmed by devex + security) + U3 test verification
- Wave 2: `toNotificationParams` unit tests + nil-param guards + broadcast delivery assertion
- Wave 3: localhost test green-when-broken fix + variadic-bool → plain bool + heartbeat doc
- Wave 4: security-verified no 401 swallowing path
- Wave 5: `translateUnknownToolError` regression test + elicitation guard branches + doc fixes

Refs #156

### DIFF
--- a/mcpcompat/client/client.go
+++ b/mcpcompat/client/client.go
@@ -51,12 +51,87 @@ type Client struct {
 	// resumes a pre-existing session (transport.WithSession) without calling
 	// Initialize. It is lazily created on the first resumed request. See resume.go.
 	resume *resumeState
+
+	// elicitationHandler, when set, is wired into the go-sdk ClientOptions so
+	// the client declares the elicitation capability at initialize and answers
+	// server->client elicitation/create requests. It mirrors mcp-go's
+	// client.WithElicitationHandler.
+	elicitationHandler ElicitationHandler
+}
+
+// ElicitationHandler handles server->client elicitation/create requests. It
+// mirrors mcp-go's client.ElicitationHandler.
+type ElicitationHandler interface {
+	Elicit(ctx context.Context, request mcp.ElicitationRequest) (*mcp.ElicitationResult, error)
+}
+
+// ElicitationHandlerFunc is an adapter to allow the use of ordinary functions
+// as elicitation handlers. It is a shim convenience adapter not present in
+// mcp-go (mcp-go's client.WithElicitationHandler accepts a plain func), kept
+// here so the shim's ElicitationHandler interface and the func form compose.
+type ElicitationHandlerFunc func(ctx context.Context, request mcp.ElicitationRequest) (*mcp.ElicitationResult, error)
+
+// Elicit calls f(ctx, request).
+func (f ElicitationHandlerFunc) Elicit(ctx context.Context, request mcp.ElicitationRequest) (*mcp.ElicitationResult, error) {
+	return f(ctx, request)
+}
+
+// ClientOption configures a Client. It is applied at construction time via
+// NewStreamableHttpClientWithOpts (the only constructor that accepts
+// ClientOptions); NewStreamableHttpClient and NewSSEMCPClient take only
+// transport-level options.
+//
+//nolint:revive // name intentionally matches mcp-go for drop-in compatibility.
+type ClientOption func(*Client)
+
+// WithElicitationHandler installs a handler for server->client
+// elicitation/create requests. The handler must be registered before Initialize
+// so it is wired into the underlying go-sdk client, which (a) declares the
+// elicitation capability at initialize time so the server's Elicit calls pass
+// the go-sdk capability gate, and (b) dispatches incoming elicitation/create
+// requests to the handler.
+//
+// Elicitation under the Streamable HTTP transport requires the client to hold
+// an open standalone SSE stream: the shim server replies with application/json
+// (JSONResponse is on by design), so the go-sdk routes a server->client
+// elicitation request made during request handling to the standalone SSE
+// stream rather than the POST response. Configure the client with
+// transport.WithContinuousListening() to open that stream; without it
+// (DisableStandaloneSSE defaults to true) elicitation cannot be delivered and
+// the server's Elicit call is rejected. See the build() doc comment in
+// mcpcompat/server/transports.go.
+//
+// It mirrors mcp-go's client.WithElicitationHandler.
+func WithElicitationHandler(h ElicitationHandler) ClientOption {
+	return func(c *Client) { c.elicitationHandler = h }
+}
+
+// applyClientOptions applies the client-level options to c.
+func applyClientOptions(c *Client, opts []ClientOption) {
+	for _, opt := range opts {
+		opt(c)
+	}
 }
 
 // NewStreamableHttpClient creates a Streamable HTTP MCP client for baseURL. Like
 // mcp-go, the returned client is not yet connected; call Start then Initialize.
+// Transport options configure the underlying Streamable HTTP transport; client
+// options (client.WithElicitationHandler, ...) configure client-level behavior.
 func NewStreamableHttpClient(baseURL string, options ...transport.StreamableHTTPCOption) (*Client, error) {
 	return &Client{streamable: transport.NewStreamableHTTP(baseURL, options...)}, nil
+}
+
+// NewStreamableHttpClientWithOpts creates a Streamable HTTP MCP client for
+// baseURL, applying transport-level options followed by client-level options.
+// It is the entry point for options that are not transport options (e.g.
+// WithElicitationHandler); NewStreamableHttpClient remains for callers that
+// only need transport options.
+func NewStreamableHttpClientWithOpts(
+	baseURL string, transportOpts []transport.StreamableHTTPCOption, clientOpts []ClientOption,
+) (*Client, error) {
+	c := &Client{streamable: transport.NewStreamableHTTP(baseURL, transportOpts...)}
+	applyClientOptions(c, clientOpts)
+	return c, nil
 }
 
 // NewSSEMCPClient creates an SSE MCP client for baseURL. The returned client is
@@ -72,6 +147,16 @@ func (*Client) Start(_ context.Context) error { return nil }
 
 // Initialize connects the underlying go-sdk client and performs the MCP
 // initialize handshake using the supplied client info and capabilities.
+//
+// LIMITATION (issue #156, item U2): if a preset session ID was supplied via
+// transport.WithSession, Initialize does NOT honor it. Initialize always
+// negotiates a fresh, server-assigned session ID: it builds a go-sdk
+// StreamableClientTransport (buildTransport) which has no preset session-ID
+// field (go-sdk has no such field in any version, verified through v1.6.1),
+// performs the initialize handshake, and then adopts the ID the server returns
+// (overwriting any value set via WithSession). To resume a pre-existing session
+// by ID, skip Initialize and use the resume path instead (see resume.go). The
+// resume path is exercised by resume_test.go.
 func (c *Client) Initialize(ctx context.Context, request mcp.InitializeRequest) (*mcp.InitializeResult, error) {
 	ctx = withErrCapture(ctx)
 	c.mu.Lock()
@@ -101,6 +186,7 @@ func (c *Client) Initialize(ctx context.Context, request mcp.InitializeRequest) 
 		opts.Capabilities = caps
 	}
 	c.installNotificationHandlers(opts)
+	c.installElicitationHandler(opts)
 
 	gc := gosdk.NewClient(impl, opts)
 
@@ -495,6 +581,46 @@ func (c *Client) installNotificationHandlers(opts *gosdk.ClientOptions) {
 	// *LoggingMessageRequest; convert them to mcp-go's notification params
 	// (level, data, logger) and dispatch.
 	opts.LoggingMessageHandler = newLoggingMessageHandler(c.dispatch)
+}
+
+// installElicitationHandler wires the user-supplied ElicitationHandler (set via
+// WithElicitationHandler) onto the go-sdk ClientOptions. The go-sdk
+// auto-declares the elicitation capability when ElicitationHandler is non-nil
+// (see Client.capabilities), so the server's Elicit calls pass the capability
+// gate; incoming elicitation/create requests are dispatched to the handler.
+//
+// The go-sdk handler receives a *gosdk.ElicitRequest and must return a
+// *gosdk.ElicitResult. The shim converts between the go-sdk and mcp-go-shaped
+// elicitation types via JSON round-trips (both encode the identical MCP wire
+// format), so a handler registered against the mcp-go ElicitationRequest/
+// ElicitationResult types works unchanged. The go-sdk validates an accepted
+// result against the requested schema itself (in (*Client).elicit), so the
+// shim does not re-validate here.
+func (c *Client) installElicitationHandler(opts *gosdk.ClientOptions) {
+	if c.elicitationHandler == nil {
+		return
+	}
+	opts.ElicitationHandler = func(ctx context.Context, req *gosdk.ElicitRequest) (*gosdk.ElicitResult, error) {
+		mreq := mcp.ElicitationRequest{}
+		if req != nil {
+			// Reconstruct the mcp-go-shaped request from the go-sdk params. The
+			// go-sdk ElicitRequest embeds its Params; a JSON round-trip maps them
+			// onto mcp-go's ElicitationParams (mode, message, requestedSchema, url,
+			// elicitationId, _meta).
+			if err := jsonConvert(req.Params, &mreq.Params); err != nil {
+				return nil, fmt.Errorf("converting elicitation request: %w", err)
+			}
+		}
+		res, err := c.elicitationHandler.Elicit(ctx, mreq)
+		if err != nil {
+			return nil, err
+		}
+		out := &gosdk.ElicitResult{}
+		if err := jsonConvert(res, out); err != nil {
+			return nil, fmt.Errorf("converting elicitation result: %w", err)
+		}
+		return out, nil
+	}
 }
 
 // dispatchFunc is the signature of Client.dispatch, factored out so the
--- a/mcpcompat/client/client.go
+++ b/mcpcompat/client/client.go
@@ -173,6 +173,23 @@ func (c *Client) Ping(ctx context.Context) error {
 	return s.Ping(ctx, nil)
 }
 
+// SetLoggingLevel sets the server's logging level. This is a renamed and
+// simplified counterpart to mcp-go's client.Client.SetLevel: rather than
+// taking an mcp.SetLevelRequest, it accepts the mcp.LoggingLevel directly.
+// The SetLevel, SetLevelRequest and SetLevelParams types from mcp-go are
+// intentionally NOT re-exported by this shim. The server only delivers
+// notifications/message notifications at or above the requested level, so this
+// must be called before the OnNotification handler will see logging
+// notifications. level is one of the MCP logging levels ("debug", "info",
+// "notice", "warning", "error", "critical", "alert", "emergency").
+func (c *Client) SetLoggingLevel(ctx context.Context, level mcp.LoggingLevel) error {
+	s, err := c.sessionFor()
+	if err != nil {
+		return err
+	}
+	return s.SetLoggingLevel(ctx, &gosdk.SetLoggingLevelParams{Level: gosdk.LoggingLevel(level)})
+}
+
 // ListTools lists the server's tools.
 func (c *Client) ListTools(ctx context.Context, request mcp.ListToolsRequest) (*mcp.ListToolsResult, error) {
 	ctx = withErrCapture(ctx)
@@ -395,34 +412,120 @@ func (c *Client) ListResourceTemplates(
 // Handlers must be registered before Initialize so they can be wired into the
 // underlying go-sdk client. The go-sdk exposes typed notification handlers
 // rather than a single catch-all, so this shim synthesizes JSONRPCNotification
-// values for the list-changed, progress and logging notifications.
+// values (method + params) for the list-changed, progress and logging
+// notifications, forwarding each to every registered handler.
+//
+// Server-initiated notifications (including the list_changed notifications,
+// which the server emits outside any in-flight request) are only delivered if
+// the client enabled continuous listening via
+// transport.WithContinuousListening(). Without it the go-sdk streamable
+// transport has no standalone SSE stream to carry such notifications and they
+// are silently dropped; no callback fires.
 func (c *Client) OnNotification(handler func(notification mcp.JSONRPCNotification)) {
 	c.notifyMu.Lock()
 	defer c.notifyMu.Unlock()
 	c.notify = append(c.notify, handler)
 }
 
-func (c *Client) dispatch(method string) {
+// dispatch fans a synthesized JSONRPCNotification out to every registered
+// OnNotification handler. params is converted into mcp-go's
+// JSONRPCNotificationParams (Meta + AdditionalFields) when non-nil, so the
+// notification carries its params as mcp-go would; nil leaves the params empty
+// (used by the list_changed notifications, which carry no params on the wire).
+func (c *Client) dispatch(method string, params any) {
 	c.notifyMu.Lock()
 	handlers := make([]func(mcp.JSONRPCNotification), len(c.notify))
 	copy(handlers, c.notify)
 	c.notifyMu.Unlock()
 	n := mcp.JSONRPCNotification{JSONRPC: mcp.JSONRPC_VERSION}
 	n.Method = method
+	if params != nil {
+		n.Params = toNotificationParams(params)
+	}
 	for _, h := range handlers {
 		h(n)
 	}
 }
 
+// toNotificationParams converts a go-sdk notification params value into the
+// mcp-go NotificationParams shape (Meta + AdditionalFields). Both encode to the
+// same MCP wire JSON, so a JSON round-trip faithfully maps the go-sdk struct's
+// fields onto mcp-go's AdditionalFields map. This mirrors the jsonConvert
+// convention used elsewhere in the shim for cross-type conversion.
+func toNotificationParams(src any) mcp.NotificationParams {
+	// Marshal the go-sdk params and unmarshal into a generic map, then split the
+	// reserved "_meta" key (if present) from the remaining fields, which become
+	// AdditionalFields.
+	b, err := json.Marshal(src)
+	if err != nil {
+		return mcp.NotificationParams{}
+	}
+	var m map[string]any
+	if err := json.Unmarshal(b, &m); err != nil {
+		return mcp.NotificationParams{}
+	}
+	out := mcp.NotificationParams{AdditionalFields: make(map[string]any, len(m))}
+	for k, v := range m {
+		if k == "_meta" {
+			if mm, ok := v.(map[string]any); ok {
+				out.Meta = mm
+			}
+			continue
+		}
+		out.AdditionalFields[k] = v
+	}
+	return out
+}
+
 func (c *Client) installNotificationHandlers(opts *gosdk.ClientOptions) {
 	opts.ToolListChangedHandler = func(_ context.Context, _ *gosdk.ToolListChangedRequest) {
-		c.dispatch("notifications/tools/list_changed")
+		c.dispatch("notifications/tools/list_changed", nil)
 	}
 	opts.PromptListChangedHandler = func(_ context.Context, _ *gosdk.PromptListChangedRequest) {
-		c.dispatch("notifications/prompts/list_changed")
+		c.dispatch("notifications/prompts/list_changed", nil)
 	}
 	opts.ResourceListChangedHandler = func(_ context.Context, _ *gosdk.ResourceListChangedRequest) {
-		c.dispatch("notifications/resources/list_changed")
+		c.dispatch("notifications/resources/list_changed", nil)
+	}
+	// Forward server->client progress notifications. go-sdk hands the params via
+	// *ProgressNotificationClientRequest; convert them to mcp-go's notification
+	// params (progressToken, progress, total, message) and dispatch.
+	opts.ProgressNotificationHandler = newProgressNotificationHandler(c.dispatch)
+	// Forward server->client logging notifications. go-sdk hands the params via
+	// *LoggingMessageRequest; convert them to mcp-go's notification params
+	// (level, data, logger) and dispatch.
+	opts.LoggingMessageHandler = newLoggingMessageHandler(c.dispatch)
+}
+
+// dispatchFunc is the signature of Client.dispatch, factored out so the
+// notification handlers can be unit-tested in isolation without a live session.
+type dispatchFunc func(method string, params any)
+
+// newProgressNotificationHandler builds the go-sdk progress-notification
+// handler that forwards notifications/progress to the supplied dispatch func.
+// If the go-sdk request or its params are nil, dispatch is called with nil
+// params (an empty NotificationParams) rather than panicking.
+func newProgressNotificationHandler(dispatch dispatchFunc) func(context.Context, *gosdk.ProgressNotificationClientRequest) {
+	return func(_ context.Context, req *gosdk.ProgressNotificationClientRequest) {
+		if req == nil || req.Params == nil {
+			dispatch("notifications/progress", nil)
+			return
+		}
+		dispatch("notifications/progress", req.Params)
+	}
+}
+
+// newLoggingMessageHandler builds the go-sdk logging-notification handler that
+// forwards notifications/message to the supplied dispatch func. If the go-sdk
+// request or its params are nil, dispatch is called with nil params rather than
+// panicking.
+func newLoggingMessageHandler(dispatch dispatchFunc) func(context.Context, *gosdk.LoggingMessageRequest) {
+	return func(_ context.Context, req *gosdk.LoggingMessageRequest) {
+		if req == nil || req.Params == nil {
+			dispatch("notifications/message", nil)
+			return
+		}
+		dispatch("notifications/message", req.Params)
 	}
 }
 
--- a/mcpcompat/client/client.go
+++ b/mcpcompat/client/client.go
@@ -262,18 +262,27 @@ func (c *Client) Ping(ctx context.Context) error {
 // SetLoggingLevel sets the server's logging level. This is a renamed and
 // simplified counterpart to mcp-go's client.Client.SetLevel: rather than
 // taking an mcp.SetLevelRequest, it accepts the mcp.LoggingLevel directly.
-// The SetLevel, SetLevelRequest and SetLevelParams types from mcp-go are
-// intentionally NOT re-exported by this shim. The server only delivers
-// notifications/message notifications at or above the requested level, so this
-// must be called before the OnNotification handler will see logging
-// notifications. level is one of the MCP logging levels ("debug", "info",
-// "notice", "warning", "error", "critical", "alert", "emergency").
+// The server only delivers notifications/message notifications at or above the
+// requested level, so this must be called before the OnNotification handler
+// will see logging notifications. level is one of the MCP logging levels
+// ("debug", "info", "notice", "warning", "error", "critical", "alert",
+// "emergency").
 func (c *Client) SetLoggingLevel(ctx context.Context, level mcp.LoggingLevel) error {
 	s, err := c.sessionFor()
 	if err != nil {
 		return err
 	}
 	return s.SetLoggingLevel(ctx, &gosdk.SetLoggingLevelParams{Level: gosdk.LoggingLevel(level)})
+}
+
+// SetLevel is a drop-in compatibility alias for mcp-go's
+// client.Client.SetLevel, forwarding to SetLoggingLevel. It is provided so
+// downstream code using the upstream idiom
+// `c.SetLevel(ctx, mcp.SetLevelRequest{...})` compiles against the shim
+// unchanged. SetLevelRequest and SetLevelParams are re-exported from the mcp
+// package for the same reason.
+func (c *Client) SetLevel(ctx context.Context, request mcp.SetLevelRequest) error {
+	return c.SetLoggingLevel(ctx, request.Params.Level)
 }
 
 // ListTools lists the server's tools.
@@ -498,8 +507,15 @@ func (c *Client) ListResourceTemplates(
 // Handlers must be registered before Initialize so they can be wired into the
 // underlying go-sdk client. The go-sdk exposes typed notification handlers
 // rather than a single catch-all, so this shim synthesizes JSONRPCNotification
-// values (method + params) for the list-changed, progress and logging
-// notifications, forwarding each to every registered handler.
+// values (method + params) for each notification type and forwards each to
+// every registered handler. The following notification types are forwarded:
+//   - notifications/tools/list_changed
+//   - notifications/prompts/list_changed
+//   - notifications/resources/list_changed
+//   - notifications/resources/updated (from resources/subscribe)
+//   - notifications/elicitation/complete (out-of-band elicitation completion)
+//   - notifications/progress (server->client progress)
+//   - notifications/message (server->client logging)
 //
 // Server-initiated notifications (including the list_changed notifications,
 // which the server emits outside any in-flight request) are only delivered if
@@ -572,6 +588,21 @@ func (c *Client) installNotificationHandlers(opts *gosdk.ClientOptions) {
 	}
 	opts.ResourceListChangedHandler = func(_ context.Context, _ *gosdk.ResourceListChangedRequest) {
 		c.dispatch("notifications/resources/list_changed", nil)
+	}
+	// Forward server->client resource-updated notifications (the result of a
+	// resources/subscribe). go-sdk hands the params via
+	// *ResourceUpdatedNotificationRequest; dispatch synthesizes the
+	// notifications/resources/updated notification with its params (uri).
+	opts.ResourceUpdatedHandler = func(_ context.Context, req *gosdk.ResourceUpdatedNotificationRequest) {
+		c.dispatch("notifications/resources/updated", req.Params)
+	}
+	// Forward server->client elicitation-complete notifications (out-of-band
+	// elicitation completion). go-sdk hands the params via
+	// *ElicitationCompleteNotificationRequest; dispatch synthesizes the
+	// notifications/elicitation/complete notification with its params
+	// (elicitationId).
+	opts.ElicitationCompleteHandler = func(_ context.Context, req *gosdk.ElicitationCompleteNotificationRequest) {
+		c.dispatch("notifications/elicitation/complete", req.Params)
 	}
 	// Forward server->client progress notifications. go-sdk hands the params via
 	// *ProgressNotificationClientRequest; convert them to mcp-go's notification
--- a/mcpcompat/client/errors.go
+++ b/mcpcompat/client/errors.go
@@ -42,6 +42,14 @@ func enrichWithResponseBody(ctx context.Context, err error) error {
 // string matching when no body was captured (e.g. a failure before any
 // response). 401 stays ErrUnauthorized so ToolHive's OAuth refresh flow still
 // triggers.
+//
+// JSON-RPC-level errors (a *jsonrpc.Error returned inside a 2xx response) that
+// carry no captured HTTP status are application-level initialize rejections, not
+// transport failures: they must NOT be reclassified as transport auth/session
+// failures based on their text (mirrors mapCallError). Connect runs initialize
+// through the same handleSend machinery as the call methods (go-sdk v1.6.1
+// client.go:271), so the same *jsonrpc.Error reaches here with no captured
+// status.
 func mapConnectError(ctx context.Context, err error) error {
 	err = enrichWithResponseBody(ctx, err)
 	h := capturedErr(ctx)
@@ -54,6 +62,14 @@ func mapConnectError(ctx context.Context, err error) error {
 			))
 		}
 		return transport.NewError(errors.Join(transport.ErrLegacySSEServer, err))
+	}
+	var wireErr *jsonrpc.Error
+	if (h == nil || h.status == 0) && errors.As(err, &wireErr) {
+		// RPC-level error with no captured HTTP status: an application-level
+		// initialize rejection (200 OK + JSON-RPC error body), not a transport
+		// failure. Surface unchanged (mirrors mapCallError) rather than
+		// string-matching it into a transport auth/legacy-SSE sentinel.
+		return err
 	}
 	return mapTransportError(ctx, err, h)
 }
--- a/mcpcompat/client/errors.go
+++ b/mcpcompat/client/errors.go
@@ -48,14 +48,14 @@ func mapConnectError(ctx context.Context, err error) error {
 	if h != nil && h.status >= 400 && h.status < 500 {
 		if h.status == http.StatusUnauthorized {
 			return transport.NewError(errors.Join(
-				&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(err)},
+				&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(ctx)},
 				transport.ErrUnauthorized,
 				err,
 			))
 		}
 		return transport.NewError(errors.Join(transport.ErrLegacySSEServer, err))
 	}
-	return mapTransportError(err, h)
+	return mapTransportError(ctx, err, h)
 }
 
 // mapCallError maps an error returned by an underlying go-sdk request call onto
@@ -85,7 +85,7 @@ func mapCallError(ctx context.Context, err error) error {
 		// RPC-level error with no captured HTTP status: surface unchanged.
 		return err
 	}
-	return mapTransportError(err, h)
+	return mapTransportError(ctx, err, h)
 }
 
 // mapTransportError inspects err and, when it recognizes an HTTP auth/session
@@ -98,42 +98,76 @@ func mapCallError(ctx context.Context, err error) error {
 // (e.g. a transport failure before any response, or a response whose body
 // capture was skipped), detection falls back to best-effort string matching.
 //
+// A captured 5xx status surfaces the error unchanged: a 5xx body containing the
+// substring "401" or "unauthorized" must NOT trigger ToolHive's OAuth refresh
+// flow (issue #156, wave 4 review finding). The string fallback only runs when
+// NO status was captured at all.
+//
 // NOTE: the go-sdk does not currently expose a typed error carrying the HTTP
 // status code, so the string-matching fallback is inherently best-effort. When
 // the pattern is not recognized the original error is returned unchanged. This is
 // the one area of the client shim where exact parity with mcp-go's OAuth flow
 // may need refinement as the go-sdk's error surface evolves.
-func mapTransportError(err error, h *errBody) error {
+func mapTransportError(ctx context.Context, err error, h *errBody) error {
 	if err == nil {
 		return nil
 	}
 	// Prefer the captured HTTP status over string matching: the status is
 	// authoritative, while body text (re-attached by enrichWithResponseBody)
 	// can legitimately contain words like "unauthorized" for non-401 responses.
-	if h != nil && h.status >= 400 && h.status < 500 {
-		switch h.status {
-		case http.StatusUnauthorized:
-			return transport.NewError(errors.Join(
-				&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(err)},
-				transport.ErrUnauthorized,
-				err,
-			))
-		case http.StatusNotFound:
-			return transport.NewError(errors.Join(transport.ErrSessionTerminated, err))
-		default:
-			// Other 4xx on a regular call are not legacy-SSE (that classification
-			// is connect-time only, in mapConnectError); surface unchanged.
-			return err
+	if h != nil && h.status >= 400 {
+		if ok, mapped := mapStatusError(ctx, err, h); ok {
+			return mapped
 		}
 	}
-	msg := strings.ToLower(err.Error())
+	// No captured status (or a captured status we don't classify, e.g. 5xx):
+	// fall back to best-effort string matching. A captured 5xx is surfaced
+	// unchanged by mapStatusError (ok=true, mapped=err), so the string fallback
+	// only runs when no status was captured at all.
+	return mapStringError(ctx, err)
+}
 
+// mapStatusError classifies an error using the captured HTTP status code. It
+// returns (ok, mapped): ok=true when the status was captured and classified
+// (including the 5xx pass-through), or ok=false when no status was captured (h
+// is nil or h.status == 0), signaling the caller to use the string fallback.
+func mapStatusError(ctx context.Context, err error, h *errBody) (bool, error) {
+	if h == nil || h.status == 0 {
+		return false, nil
+	}
+	if h.status < 400 {
+		return false, nil
+	}
+	if h.status >= 500 {
+		// A captured 5xx must not fall through to string matching: a 5xx body
+		// containing "401"/"unauthorized" would falsely trigger OAuth refresh.
+		return true, err
+	}
+	switch h.status {
+	case http.StatusUnauthorized:
+		return true, transport.NewError(errors.Join(
+			&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(ctx)},
+			transport.ErrUnauthorized,
+			err,
+		))
+	case http.StatusNotFound:
+		return true, transport.NewError(errors.Join(transport.ErrSessionTerminated, err))
+	default:
+		// Other 4xx on a regular call are not legacy-SSE (that classification
+		// is connect-time only, in mapConnectError); surface unchanged.
+		return true, err
+	}
+}
+
+// mapStringError classifies an error using best-effort string matching on its
+// message. This is the fallback path used when no HTTP status was captured
+// (e.g. a transport failure before any response).
+func mapStringError(ctx context.Context, err error) error {
+	msg := strings.ToLower(err.Error())
 	switch {
 	case strings.Contains(msg, "401") || strings.Contains(msg, "unauthorized"):
-		// 401: ToolHive checks both ErrAuthorizationRequired (and As
-		// *AuthorizationRequiredError / *transport.Error) and ErrUnauthorized.
 		return transport.NewError(errors.Join(
-			&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(err)},
+			&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(ctx)},
 			transport.ErrUnauthorized,
 			err,
 		))
@@ -146,9 +180,16 @@ func mapTransportError(err error, h *errBody) error {
 	}
 }
 
-// extractResourceMetadataURL is a placeholder for parsing the RFC 9728
-// resource_metadata parameter out of a WWW-Authenticate header. The go-sdk does
-// not surface the header on the error today, so this returns empty for now.
-func extractResourceMetadataURL(_ error) string {
-	return ""
+// extractResourceMetadataURL parses the RFC 9728 §5.1 resource_metadata
+// parameter from the WWW-Authenticate header captured by the shim's own
+// headerRoundTripper (see captureErrorBody). The go-sdk does not surface the
+// header on its errors, but the shim's RoundTripper has direct access to
+// resp.Header and captures it into the per-call errBody holder. Returns empty
+// when no header was captured or no resource_metadata parameter is present.
+func extractResourceMetadataURL(ctx context.Context) string {
+	h := capturedErr(ctx)
+	if h == nil {
+		return ""
+	}
+	return extractResourceMetadataURLFromHeaders(h.wwwAuthHdrs)
 }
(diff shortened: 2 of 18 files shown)


mcpcompat/server/transports.go
```diff
@@ -130,24 +136,84 @@ func WithEndpointPath(endpointPath string) StreamableHTTPOption {
...
 
-// WithHeartbeatInterval sets the keep-alive ping interval.
+// WithHeartbeatInterval sets the keep-alive ping interval. This option is now
+// LIVE (it was previously a no-op): it is wired to go-sdk's
+// ServerOptions.KeepAlive, so the SDK pings the peer at this interval and
+// closes the session when a ping goes unanswered. Set a small interval only if
+// you want unanswered pings to terminate the session. Applies to the Streamable
+// HTTP transport only; stdio and SSE pass 0 (no keep-alive), matching mcp-go.
 func WithHeartbeatInterval(interval time.Duration) StreamableHTTPOption {
 	return func(s *StreamableHTTPServer) { s.heartbeat = interval }
```
JAORMX, line 188: **CRITICAL — this evicts healthy sessions; vMCP will lose every JSON-only client session ~30s after creation.** Failure scenario: go-sdk's KeepAlive (shared.go:593-637) sends a ping *request* each interval and `session.Close()`s on failure. This handler forces `JSONResponse: true`, so a server→client request outside an in-flight POST routes to the standalone SSE stream (go-sdk streamable.go:1410-1412); a client that never opened a GET stream — the shim client's own default (`DisableStandaloneSSE: true`) — has that stream unconnected, the write is rejected immediately ("stream not connected"), the ping errors, and the session is closed on the first tick. toolhive's vMCP sets `WithHeartbeatInterval(30s)` unconditionally (pkg/vmcp/server/server.go:505), so after #5729 bumps to this release, every session from a JSON-only client dies in ~30s. mcp-go's heartbeat was passive: pings written only to an existing GET stream, never awaited, never terminating (mcp-go streamable_http.go:827-856). `TestHeartbeat_WiredToKeepAlive` enshrines the failure as expected — its "unresponsive" client is just a normal JSON-only client that could never receive the ping; there is no test that an idle healthy client *survives* the heartbeat, because under this wiring it can't. Fix options: passive shim-side keepalive (the issue's original SSE-comment design), ping only sessions with a connected standalone stream, or gate the toolhive#5729 merge on removing `WithHeartbeatInterval` from vMCP. Do not ship both as-is.

mcpcompat/server/server.go
```diff
@@ -533,16 +710,56 @@ func toGoSDKTool(t mcp.Tool) (*gosdk.Tool, error) {
...
+// passes through here (e.g. $ref, oneOf, a boolean schema, or an object schema
+// with type omitted) will panic at registration time. We pass such schemas
+// through verbatim (rather than silently rewriting them) so callers surface the
+// incompatibility rather than silently receiving a different schema than the one
+// they declared; mcp-go performed no validation here either. The two registration
+// paths handle the panic differently: the per-session path
+// (MCPServer.addSessionTool, used by syncSessionTools) recovers and skips the
+// offending tool so one bad overlay schema cannot crash a live session; the
+// global path (buildServer) lets the panic surface as a construction-time error
+// (a 500 at handler build).
```
JAORMX: **HIGH — a single global tool with a non-object schema permanently bricks the server (empirical repro).** This comment's claim that the global path surfaces "a construction-time error (a 500 at handler build)" is wrong: `buildServer` does not recover the AddTool panic, and `StreamableHTTPServer.build()` runs it inside `sync.Once.Do` (transports.go:217-247). `Once.Do` marks itself done even when f panics, so after the first request panics, `buildErr` is nil and `handler` is nil — every subsequent request nil-panics at `s.handler.ServeHTTP` (transports.go:326). Reproduced: register one global tool via `NewToolWithRawSchema` with `{"$ref":...}` → request 1 = connection EOF, request 2+ = `http: panic serving ... nil pointer dereference`, forever. `ServeStdio` (transports.go:62-68) crashes the bridge process at startup instead. go-sdk panics for any top-level type ≠ "object" (server.go:247-259), which includes the spec-loose but common `"type": ["object", "null"]` and type-omitted object schemas — under mcp-go these were served verbatim and fully callable; under v0.0.27 they were rewritten but callable; now they brick the server (global) or silently drop the tool (per-session), both worse. Minimum fix: recover in `buildServer` and return the panic as an error into `buildErr`; better: normalize non-object schemas while preserving `properties`/`required` instead of dropping. Either way, add a global-path test — only the per-session recovery is covered.

mcpcompat/server/transports.go
```diff
@@ -218,6 +299,30 @@ func (s *StreamableHTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+	// every request. When a shared SessionIdManager is configured, validate local
+	// sessions on every request too, so a session terminated in the shared store
+	// (e.g. via a DELETE on another replica) is rejected here and its local
+	// bookkeeping dropped, rather than being served forever from the local
+	// handler. The single-replica fast-path (no manager) is preserved: there is
+	// nothing to validate against, so local sessions route straight through.
+	if sid := r.Header.Get("Mcp-Session-Id"); sid != "" && s.sessionIDMgr != nil && s.mcp.isLocalSession(sid) {
+		isTerminated, err := s.sessionIDMgr.Validate(sid)
+		if err != nil {
+			s.mcp.forgetSession(sid)
```
JAORMX: **MEDIUM — a transient Validate error permanently downgrades a local session into split-brain.** Failure scenario: Redis blips once → `Validate` returns err → `forgetSession(sid)` drops the clientSession (per-session tool overlay included) and the local marker, but the go-sdk StreamableHTTPHandler still holds the live session. The client's retry now takes the `!isLocalSession` branch into `serveRehydrated`, which builds a *second* go-sdk session with the same ID — two live sessions, one ID, one pod; the client's standalone GET stream stays attached to the old one, so notifications/elicitations go nowhere. mcp-go rejected the request and kept state. Only forget on `isTerminated == true`; on err, just reject the request. This path has no test.

mcpcompat/client/errors.go
```diff
@@ -32,41 +33,99 @@ func enrichWithResponseBody(ctx context.Context, err error) error {
...
+// SSE-only server. We restore that behavior here using the captured HTTP status
+// (set by captureErrorBody for non-2xx responses), falling back to best-effort
+// string matching when no body was captured (e.g. a failure before any
+// response). 401 stays ErrUnauthorized so ToolHive's OAuth refresh flow still
+// triggers.
 func mapConnectError(ctx context.Context, err error) error {
-	return mapTransportError(enrichWithResponseBody(ctx, err))
+	err = enrichWithResponseBody(ctx, err)
+	h := capturedErr(ctx)
+	if h != nil && h.status >= 400 && h.status < 500 {
```
JAORMX, line 56: **MEDIUM — status-driven classification silently disabled for empty-body 4xx.** `captureErrorBody` (errorbody.go, not in this diff) records `h.status` only when the body is non-empty (`if err == nil && len(data) > 0`). A 4xx with an empty body — common for 401 and 404 — leaves `h.status == 0`, this branch never fires, and classification falls to the string fallback, which does not map a bare "400 Bad Request"/"404 Not Found" on initialize to `ErrLegacySSEServer` (mcp-go classified any 4xx from the status line alone). Every wave-4 test seeds a non-empty body, so the hole is untested. One-line fix: record `h.status` regardless of body length, and add an empty-body case to `TestMapConnectError_4xxOnInitialize_LegacySSE`.

mcpcompat/server/server.go
```diff
@@ -128,46 +156,60 @@ type MCPServer struct {
 	// in registerAndSync (which only fires on this instance's initialize path).
 	localSessions sync.Map // sessionID -> struct{}
```
JAORMX, line 174: **LOW — unbounded growth: entries here (and in `sessions`) are never removed when go-sdk closes a session.** Only DELETE and the Validate-failure path call `forgetSession`; sessions closed by keepalive eviction or a vanished client leak their `localSessions` + `sessions` entries for the life of the process, and `SendNotificationToAllClients` keeps iterating the corpses. On a long-lived vMCP pod this grows without bound — and the heartbeat issue above turns it into mass leakage. `sessions` predates this PR; `localSessions` is new here. Worth a cleanup hook (e.g. reap on Validate 404 from the go-sdk handler, or periodically drop entries whose goSession is closed).

mcpcompat/server/transports.go
```diff
@@ -182,16 +257,22 @@ func (s *StreamableHTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+	// middleware reads it back via req.GetExtra().Header to bridge the values.
+	// Keying by nonce (not session ID) avoids the race where concurrent POSTs
+	// on the same session clobber each other's context (issue #156, item U3).
+	// The initialize POST (no session yet) carries no per-request values the
+	// handler needs; only non-initialize POSTs need bridging, but bridging
+	// initialize is harmless.
+	nonce := crand.Text()
+	s.mcp.setPendingRequestContext(r.Context(), nonce)
+	r.Header.Set(reqNonceHeader, nonce)
+	defer s.mcp.clearPendingRequestContext(nonce)
```
JAORMX, line 307: **LOW — the bridge can miss notification POSTs.** A notification-only POST gets its 202 immediately, so this deferred clear can run before the session goroutine dispatches the message; the middleware then finds no entry and falls back to the initialize-frozen session context — per-request values (identity, audit, telemetry) are silently absent for notification handling. No cross-request bleed (calls block in servePOST until the response, so their entries are always present), and nothing depends on it today (`registerAndSync` runs on the initialize *call*, not on notifications/initialized), but worth a sentence in the `pendingReqCtx` doc so the next reader doesn't assume the bridge covers notifications.

JAORMX (commented):
(Submitted as COMMENT — GitHub rejects REQUEST_CHANGES on one's own PR; treat this as request-changes.) Verified each fix against go-sdk v1.6.1 and mcp-go v0.55.1 sources; U1/U3/U9/5/6 and the 4/U7 core are correct. Four issues must change first — see inline comments (heartbeat session eviction, global schema panic → poisoned sync.Once, Validate-error forgetSession, empty-body 4xx capture). Separately, the "zero functional loss" claim needs reconciling: `notifications/tools/list_changed` and `notifications/message` are still lost end-to-end through ToolHive's stdio bridge — the bridge never mutates its feature set (so go-sdk never auto-emits) and doesn't call `WithLogging` (so clients never setLevel and `ServerSession.Log` drops everything; the tests only pass because they call `SetLoggingLevel` first). Those need toolhive-side companions to #5729: `WithLogging()` + feature re-sync on upstream list_changed in the bridge, and `WithPageSize` in vMCP.

mcpcompat/server/transports.go
JAORMX: Addressed in `d898410`. Summary:

**1 (CRITICAL — heartbeat eviction):** Reverted the `ServerOptions.KeepAlive` wiring. `WithHeartbeatInterval` is now a documented no-op — go-sdk's active ping kills JSON-only sessions (no GET stream → ping rejected → `session.Close()` on first tick), incompatible with JSONResponse mode. Passive SSE-comment design is the correct approach and is left as `TODO(issue #156)`. Replaced `TestHeartbeat_WiredToKeepAlive` with `TestHeartbeat_NoOpDoesNotEvictHealthySession` — an idle healthy client survives past the heartbeat interval.

**2 (HIGH — global schema panic poisons sync.Once):** `buildServer` now recovers global `AddTool` panics via `addGlobalTool` and returns the panic as `buildErr`, so `sync.Once` is properly poisoned with an error (not nil-error-nil-handler). Also improved `normalizeObjectSchema`: type-omitted object schemas with `properties`/`required` get `type:"object"` added (callable, matching mcp-go); truly non-object schemas pass through and are caught by the recover. Tests: `TestGlobalTool_NonObjectSchema_DoesNotPoisonServer` (first and second request get clean 500, not nil panic), `TestGlobalTool_TypeOmittedObjectSchema_Callable`.

**3 (MEDIUM — transient Validate → split-brain):** `forgetSession`/`deleteRehydrated` now called only on genuine termination (`isTerminated == true`). On a transient Validate error, the request is rejected (503) WITHOUT dropping local state — the go-sdk session stays alive, client retries, Redis recovers. Test: `TestLocalSession_TransientValidateError_DoesNotForgetSession` (flaky manager flakes once → 503, then session still local and served on retry).

**4 (MEDIUM — empty-body 4xx):** `captureErrorBody` now records `h.status = resp.StatusCode` unconditionally (before reading the body), regardless of body length. Empty-body 401/404 are now classified by status. Added empty-body cases to `TestMapCallError_Transport401` and `TestMapConnectError_4xxOnInitialize_LegacySSE`.

**5 (LOW — unbounded session growth):** Documented the gap on `sessions`/`localSessions` — reaping closed sessions is needed and tracked separately.

**6 (LOW — bridge misses notification POSTs):** Documented at `defer clearPendingRequestContext` — the bridge covers request/response calls (which block in `servePOST`) but not notification-only POSTs (immediate 202 → deferred clear can race dispatch), pending go-sdk per-request context propagation.

**Review summary (zero functional loss):** Acknowledged — `notifications/tools/list_changed` and `notifications/message` require toolhive-side companions to #5729 (`WithLogging()` + feature re-sync in the bridge, `WithPageSize` in vMCP). The shim now provides the surface (handlers, `SetLoggingLevel`, `WithPageSize`); the bridge-side wiring is toolhive's to consume.

Also fixed a pre-existing data race exposed by the timing changes: `clientSession.owner` was a plain `*MCPServer` raced between `SetSessionTools` (before-hook goroutine) and `registerAndSync` (initialized callback). Converted to `atomic.Pointer[MCPServer]`, matching the existing `boundServer`/`goSession` pattern. 10× `-race` clean.

mcpcompat/server/server.go
JAORMX: Addressed in `d898410` — see reply on the first comment for full details. Global `AddTool` panics are now recovered in `buildServer` via `addGlobalTool`, returning the panic as `buildErr` so `sync.Once` is properly poisoned. `normalizeObjectSchema` also improved: type-omitted object schemas with `properties`/`required` get `type:"object"` added (callable). Tests: `TestGlobalTool_NonObjectSchema_DoesNotPoisonServer`, `TestGlobalTool_TypeOmittedObjectSchema_Callable`.

mcpcompat/server/transports.go
JAORMX: Addressed in `d898410` — `forgetSession`/`deleteRehydrated` now called only on `isTerminated == true`. On transient Validate error, request is rejected (503) without dropping local state, preventing the split-brain. Test: `TestLocalSession_TransientValidateError_DoesNotForgetSession`.

mcpcompat/client/errors.go
JAORMX: Addressed in `d898410` — `captureErrorBody` now records `h.status = resp.StatusCode` unconditionally (before reading the body), regardless of body length. Empty-body 4xx are now classified by status. Added empty-body cases to `TestMapCallError_Transport401` and `TestMapConnectError_4xxOnInitialize_LegacySSE`.

mcpcompat/server/server.go
JAORMX: Addressed in `d898410` — documented the gap on `sessions`/`localSessions` (reaping closed sessions needed, tracked separately). Also fixed a pre-existing data race on `clientSession.owner` (converted to `atomic.Pointer[MCPServer]`, matching the `boundServer`/`goSession` pattern) that was exposed by the timing changes.

mcpcompat/server/transports.go
JAORMX: Addressed in `d898410` — documented at the `defer clearPendingRequestContext` site that the bridge covers request/response calls (which block in `servePOST` until the response) but not notification-only POSTs (immediate 202 → deferred clear can race dispatch), pending go-sdk per-request context propagation.

mcpcompat/client/errors.go
```diff
@@ -32,41 +33,99 @@ func enrichWithResponseBody(ctx context.Context, err error) error {
...
+// carry no captured HTTP status are application errors, not transport failures:
+// they must NOT be reclassified as transport auth/session failures based on
+// their text — otherwise a tool error whose message contains "unauthorized"
+// would wrongly trigger ToolHive's OAuth refresh flow. Only errors that look
+// like transport-level HTTP failures (a captured non-2xx status, or a plain
+// non-RPC error from the transport) are passed to mapTransportError.
 func mapCallError(ctx context.Context, err error) error {
 	if err == nil {
 		return nil
 	}
```
jhrozek, line 93: Minor / possible functional gap: `extractResourceMetadataURL` (line 152) is a hard-coded empty stub returning `""`, so `AuthorizationRequiredError.ResourceMetadataURL` is always empty. The doc comment blames the go-sdk ("does not surface the header on the error today"), but the shim's own `headerRoundTripper` (client.go:765-785) has direct access to `resp.Header` and already captures the status/body into the `errBody` holder (errorbody.go). It could equally capture `resp.Header.Get("WWW-Authenticate")` and parse the RFC 9728 `resource_metadata` parameter, so this isn't actually blocked by the go-sdk.

If ToolHive's OAuth flow consumes `ResourceMetadataURL` for discovery, this is a real functional loss vs mcp-go rather than a documentation-only gap. If ToolHive only branches on `errors.Is(ErrAuthorizationRequired)` / `errors.As(*AuthorizationRequiredError)` (as the test at errors_internal_test.go:26-27 suggests), the empty field is harmless. Worth confirming which it is before merge; if the former, capturing the header in `captureErrorBody` is a small change.

mcpcompat/client/errors.go
```diff
@@ -32,41 +33,99 @@ func enrichWithResponseBody(ctx context.Context, err error) error {
...
+			))
+		case http.StatusNotFound:
+			return transport.NewError(errors.Join(transport.ErrSessionTerminated, err))
+		default:
+			// Other 4xx on a regular call are not legacy-SSE (that classification
+			// is connect-time only, in mapConnectError); surface unchanged.
+			return err
+		}
+	}
 	msg := strings.ToLower(err.Error())
```
jhrozek: Residual edge case from wave 4: the status-driven classification at line 113 only covers `h.status >= 400 && h.status < 500`, so a genuine 5xx (captured status 500-599) with a body containing the substring "401" or "unauthorized" still falls through to this string branch and would falsely trigger ToolHive's OAuth refresh flow.

Low likelihood (5xx bodies rarely contain those tokens), and the comment at lines 100-105 already flags the string fallback as best-effort pending upstream typed HTTP-status errors. But if the goal of wave 4 was to eliminate false-positive OAuth refreshes, the 5xx path is still open. A cheap belt-and-braces fix: only run the string fallback when `h == nil || h.status == 0` (i.e. no captured status at all), and return `err` unchanged when a 5xx status was captured.

mcpcompat/server/transports.go
```diff
@@ -218,6 +320,52 @@ func (s *StreamableHTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+	// we called forgetSession here, the live go-sdk session would remain in the
+	// handler's map while the local marker is gone → the client's retry takes the
+	// !isLocalSession branch → serveRehydrated builds a SECOND go-sdk session with
+	// the same ID → split-brain (two live sessions, one ID, one pod; the standalone
+	// GET stream stays on the old one). So on a non-terminated Validate error we
+	// only reject THIS request (503, retry-able) and leave the local session
+	// intact; if the shared store recovers, the next Validate succeeds and the
+	// request is served normally. Only a genuine termination (isTerminated) drops
+	// local bookkeeping.
+	if sid := r.Header.Get("Mcp-Session-Id"); sid != "" && s.sessionIDMgr != nil && s.mcp.isLocalSession(sid) {
```
jhrozek, line 354: Inconsistency with the rehydrated-session path in `serveRehydrated` (transports.go:376-405): this local-session block deliberately returns 503 and preserves local state on a transient Validate error (non-terminated, e.g. a Redis blip), with a detailed comment explaining why dropping causes split-brain. But `serveRehydrated` does the opposite — a transient Validate error there drops the cached reconstruction (`deleteRehydrated`) and returns 404 "Invalid session ID".

For a rehydrated session dropping is arguably safer (rebuild is cheap and idempotent, so a stale reconstruction is correctly discarded), so this may be intentional. But the 404 status + "Invalid session ID" body for a transient error is a sharper signal than the situation warrants, and a client may treat it as a hard termination and re-initialize. Consider 503 in `serveRehydrated`'s transient-error branch too, matching this local path's handling.

mcpcompat/server/transports.go
```diff
@@ -292,7 +440,7 @@ func (s *StreamableHTTPServer) rehydrate(r *http.Request, sid string) (*rehydrat
 		return rt, nil
 	}
 
-	srv, err := s.mcp.buildServer(nil)
+	srv, err := s.mcp.buildServer(nil, 0)
```
jhrozek, line 490: Minor scalability: `rehydrate` holds `rehydratedMu` across `buildServer` + `srv.Connect`. `buildServer` copies the tool/resource/prompt maps under `s.mu` (quick), but `Connect` can block on transport setup, so concurrent rehydrate calls for *different* session IDs serialize on this lock. The double-check at line 439 makes same-sid serialization correct/desired; the cross-sid serialization is the incidental cost.

Low impact in practice (cross-replica rehydration is rare and the first request per sid is the only one that builds), so not worth restructuring now — but if rehydration volume ever rises, building the session outside the lock and only inserting under it would remove the serialization.

mcpcompat/server/server.go
```diff
@@ -533,16 +769,79 @@ func toGoSDKTool(t mcp.Tool) (*gosdk.Tool, error) {
...
+			if _, hasReq := s["required"]; hasReq {
+				return withTypeObject(s)
+			}
+			return map[string]any{schemaTypeKey: schemaTypeObject}
 		}
+		// No "type" key at all: if the schema declares properties or required,
+		// treat it as an object schema with type omitted (spec-loose but common)
+		// and add type:"object" so go-sdk's AddTool accepts it. This matches
+		// mcp-go, which served such schemas verbatim and callable.
+		if _, hasType := s[schemaTypeKey]; !hasType {
```
jhrozek: Nit: the `if t, ok := s[schemaTypeKey].(string); ok && t == ""` branch above (line 798) and this `if _, hasType := s[schemaTypeKey]; !hasType` branch both collapse to "if properties or required present → withTypeObject, else bare object". The two could be unified into a single `typeIsMissing := !hasType || t == ""` check, which would make the symmetry between "empty type" and "no type key" explicit. Pure readability suggestion; current code is correct and well-commented.

jhrozek (commented):
Reviewed the full diff (17 files, ~3.1k LOC) plus ran `go build`, `go vet`, `go test -race ./mcpcompat/...`, `golangci-lint`, and `task license-check` — all green. The PR is high quality: the per-request nonce context bridge (U3), the 401-string-scoping to transport-level errors (wave 4), the capability-flag wiring (U1), and the schema-normalization narrowing (U6) are all correctly reasoned and well-documented. Findings below are all non-blocking nits/minor gaps; none block merge.

One additional note that I couldn't anchor inline (the relevant code is pre-existing, not in this PR's diff): `newClientSession` (mcpcompat/server/session.go:98-110) spawns a drain goroutine `for range cs.notifCh` that never terminates and is never stopped on `forgetSession`. So every `clientSession` ever created leaks a goroutine for the process lifetime. This is distinct from the documented `sessions`/`localSessions` map-growth gap (server.go:148-174): even if the planned reaper drops map entries, the goroutine stays alive because nothing closes `notifCh`. If/when the reaper lands, closing `notifCh` in `forgetSession` (session.go:316) would let the drain goroutine exit — but it needs a `sync.Once` or `LoadAndDelete` guard against double-close, since `sessionFor` can re-create an entry for the same ID. Not a blocker for this PR since it's pre-existing, but worth tracking alongside the reaper work.

jhrozek (commented):
Review of the mcpcompat functional-gaps changes. Build, `go test -race ./mcpcompat/...`, and golangci-lint are all green, and the PR is unusually well-documented. The inline comments below are gaps/edge-cases worth a look before merge; my one blocking concern is the `mapConnectError` asymmetry (first comment).

**Overall: request a look at the first inline comment before merge; the rest are non-blocking (a question, a leak worth folding in, and follow-up/doc notes).**

---

**Additional non-inline note — pre-existing goroutine leak (not introduced by this PR, but the fix would fit naturally here):**

`newClientSession` (server/session.go:98-110) spawns `go func(){ for range cs.notifCh }()`, and `notifCh` is never closed anywhere in the package. `forgetSession` (session.go:316-319) drops the map entries but doesn't close the channel, so the drain goroutine — plus the `clientSession` it references — leaks for the process lifetime, one per session ever created. This compounds the documented finding-5 (unbounded `sessions`/`localSessions` growth). Both the drain goroutine and `forgetSession` predate this PR (introduced in `bcdfd8b`, already on `main`), so it's strictly out of scope — flagging because `forgetSession` is the natural teardown point to close `notifCh` (via `sessions.LoadAndDelete` so a DELETE-then-Validate double call can't double-close), and this PR is in the neighborhood. Fine to defer to a follow-up.

_Skipped as INFO-only: (a) logging capability advertised only with `WithLogging` — matches mcp-go and delivery isn't capability-gated, just worth confirming log-emitting servers call `WithLogging`; (b) errorbody.go:72-78 comment says it "restores" resp.Body on a second error response but installs an empty one — traced unreachable in the current Connect flow._

mcpcompat/client/errors.go
```diff
@@ -32,41 +33,99 @@ func enrichWithResponseBody(ctx context.Context, err error) error {
...
+		if h.status == http.StatusUnauthorized {
+			return transport.NewError(errors.Join(
+				&transport.AuthorizationRequiredError{ResourceMetadataURL: extractResourceMetadataURL(err)},
+				transport.ErrUnauthorized,
+				err,
+			))
+		}
+		return transport.NewError(errors.Join(transport.ErrLegacySSEServer, err))
+	}
+	return mapTransportError(err, h)
```
jhrozek: **[Blocking] `mapConnectError` is missing the JSON-RPC guard that `mapCallError` has, reintroducing the false-positive the guard was added to fix — on the connect/initialize path.**

The guard added to `mapCallError` (errors.go:84-87) surfaces an application-level `*jsonrpc.Error` (no captured HTTP status) unchanged, so a tool error whose text contains "unauthorized"/"401" isn't misclassified as a transport auth failure. `mapConnectError` has no equivalent check — it falls straight to `mapTransportError`'s string-matcher when `h.status == 0`.

`Client.Connect` runs `initialize` through the **same** `handleSend` machinery as the call methods (verified in go-sdk v1.6.1: `Connect` -> `handleSend[*InitializeResult]` at client.go:271, same path as `CallTool` etc.). So a server that rejects `initialize` with a JSON-RPC-level error (200 OK + error body) whose message contains auth-flavored text gets misclassified as `ErrUnauthorized`/`ErrLegacySSEServer` here — wrongly triggering ToolHive's OAuth-refresh or legacy-SSE fallback. That's the exact class this PR fixed on the call path but not the connect path. (HTTP 4xx rejections are already handled correctly by the status branch above; this only concerns app-level JSON-RPC errors.)

Suggested symmetric fix (`jsonrpc` is already imported in this file):

```suggestion
	if h == nil || h.status == 0 {
		var wireErr *jsonrpc.Error
		if errors.As(err, &wireErr) {
			// RPC-level error with no captured HTTP status: an application-level
			// initialize rejection, not a transport failure. Surface unchanged
			// (mirrors mapCallError) rather than string-matching it into a
			// transport auth/legacy-SSE sentinel.
			return err
		}
	}
	return mapTransportError(err, h)
```

Worth adding a `TestMapConnectError_DoesNotMisclassifyJSONRPCErrors` mirroring the existing `mapCallError` test.

mcpcompat/client/client.go
```diff
@@ -173,6 +259,23 @@ func (c *Client) Ping(ctx context.Context) error {
...
+// SetLoggingLevel sets the server's logging level. This is a renamed and
+// simplified counterpart to mcp-go's client.Client.SetLevel: rather than
+// taking an mcp.SetLevelRequest, it accepts the mcp.LoggingLevel directly.
+// The SetLevel, SetLevelRequest and SetLevelParams types from mcp-go are
+// intentionally NOT re-exported by this shim. The server only delivers
+// notifications/message notifications at or above the requested level, so this
+// must be called before the OnNotification handler will see logging
+// notifications. level is one of the MCP logging levels ("debug", "info",
+// "notice", "warning", "error", "critical", "alert", "emergency").
+func (c *Client) SetLoggingLevel(ctx context.Context, level mcp.LoggingLevel) error {
```
jhrozek, line 270: **[Question] Is dropping `SetLevel` intentional, and does any importer rely on it?**

mcp-go's `client.MCPClient` exposes `SetLevel(ctx, mcp.SetLevelRequest) error`; this shim provides the renamed `SetLoggingLevel(ctx, mcp.LoggingLevel)` and deliberately doesn't re-export `SetLevel`/`SetLevelRequest`/`SetLevelParams`. Any downstream code using the upstream idiom `c.SetLevel(ctx, mcp.SetLevelRequest{...})` will fail to compile against the shim, which cuts against the drop-in contract.

This is a *new* method (nothing regressed), so it only bites if a real caller uses `SetLevel`. If none do, documenting the divergence is fine. If any do, a ~10-line compat layer restores parity while keeping the ergonomic method:

```go
// alias.go: SetLevelRequest = mcpgo.SetLevelRequest; SetLevelParams = mcpgo.SetLevelParams
func (c *Client) SetLevel(ctx context.Context, request mcp.SetLevelRequest) error {
    return c.SetLoggingLevel(ctx, request.Params.Level)
}
```

Could you confirm which way you'd like to go?

mcpcompat/client/client.go
```diff
@@ -395,34 +498,160 @@ func (c *Client) ListResourceTemplates(
...
+		c.dispatch("notifications/resources/list_changed", nil)
+	}
+	// Forward server->client progress notifications. go-sdk hands the params via
+	// *ProgressNotificationClientRequest; convert them to mcp-go's notification
+	// params (progressToken, progress, total, message) and dispatch.
+	opts.ProgressNotificationHandler = newProgressNotificationHandler(c.dispatch)
+	// Forward server->client logging notifications. go-sdk hands the params via
+	// *LoggingMessageRequest; convert them to mcp-go's notification params
+	// (level, data, logger) and dispatch.
+	opts.LoggingMessageHandler = newLoggingMessageHandler(c.dispatch)
```
jhrozek, line 614: **[Compat gap] `OnNotification` silently drops `notifications/resources/updated` (and `notifications/elicitation/complete`).**

mcp-go's `OnNotification` was a true catch-all; this shim wires only 5 typed go-sdk handlers here. go-sdk also exposes `ResourceUpdatedHandler` (client.go:154) and `ElicitationCompleteHandler` (149), left nil — so a caller doing `resources/subscribe` and waiting for `notifications/resources/updated` via `OnNotification` receives nothing, with no error and no documented gap.

Suggest wiring `opts.ResourceUpdatedHandler = func(_ context.Context, req *gosdk.ResourceUpdatedNotificationRequest){ c.dispatch("notifications/resources/updated", reqParams) }` (the `dispatch`/`toNotificationParams` machinery already exists, mirroring the progress/logging handlers just above), and either wiring `ElicitationCompleteHandler` or documenting it as unsupported in the `OnNotification` doc comment.

mcpcompat/server/transports.go
```diff
@@ -130,44 +136,122 @@ func WithEndpointPath(endpointPath string) StreamableHTTPOption {
...
+// JSON-only client (no GET stream, the shim client's own default) has that
+// stream unconnected → ping rejected → session closed on the first tick. That
+// would evict every healthy JSON-only client session at the interval. mcp-go's
+// heartbeat was passive (pings written only to an existing GET stream, never
+// awaited, never terminating).
+//
+// TODO(issue #156): implement a passive keep-alive (SSE comment injection via
+// a ResponseWriter wrapper around the go-sdk handler) matching mcp-go's design
+// and wire this option to it. Until then the value is stored but unused.
 func WithHeartbeatInterval(interval time.Duration) StreamableHTTPOption {
```
jhrozek, line 187: **[Follow-up] No keep-alive at all, while elicitation + all server->client notifications now structurally depend on the client's idle standalone SSE stream staying open.**

The rationale for *not* wiring go-sdk's active-ping `KeepAlive` (it would evict JSON-only sessions on the first tick) is sound. But the net effect is zero keep-alive, whereas mcp-go emitted passive SSE pings. Because elicitation and list-changed/progress/logging require the client's `WithContinuousListening` GET stream, any intermediary (LB, reverse proxy, ToolHive's own proxy) will reap that idle stream on timeout — after which elicitation fails (`ErrRejected`) and server-initiated notifications are silently dropped, with no reconnect. It also feeds the finding-5 unbounded-session growth (abandoned sessions are never reaped).

The TODO'd passive keep-alive (SSE-comment injection via a ResponseWriter wrapper) is more load-bearing than "nice to have" given the notification/elicitation paths depend on the stream. Reasonable as a tracked follow-up rather than in-PR, but worth prioritizing before internet-facing vMCP use; a max-session cap / idle sweep would also help.

mcpcompat/server/transports.go
```diff
@@ -218,6 +320,52 @@ func (s *StreamableHTTPServer) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
+				s.mcp.forgetSession(sid)
+				s.deleteRehydrated(sid)
+				http.Error(w, "Session terminated", http.StatusNotFound)
+				return
+			}
+			// Transient error (e.g. shared-store blip): reject this request
+			// without dropping local state. The local go-sdk session stays
+			// alive; the client retries and, once Validate succeeds again, is
+			// served normally rather than routed through rehydration.
+			http.Error(w, "session validation unavailable", http.StatusServiceUnavailable)
```
jhrozek, line 379: **[Doc note, non-blocking] 503 here diverges from mcp-go's 404 on a Validate error, which changes client re-init semantics.**

mcp-go returned 404 on *any* `Validate` error; the shim returns 503 on a non-terminated (transient) error and reserves 404 for genuine termination. Not dropping local state on a transient blip is the right call (it avoids the split-brain this PR guards against) — but 503 is unmapped by the client classifier, so `errors.Is(err, ErrSessionTerminated)` is false and a caller that keyed re-initialization off mcp-go's 404 -> `ErrSessionTerminated` path won't re-init; it just sees a generic retryable error. Worth an explicit compat-note comment so callers relying on that path are aware.

JAORMX:
## Wave 6 — addresses jhrozek's review findings

Pushed `eb17064`, all findings addressed. `task lint` (0 issues), `task test` (-race, all green), `task license-check` (clean).

### 1. 5xx false-positive OAuth refresh (MEDIUM — `errors.go:129`)

**Fixed.** `mapTransportError` now surfaces a captured 5xx unchanged — the string fallback only runs when no status was captured at all (`h == nil || h.status == 0`). A 5xx body containing "401"/"unauthorized" no longer triggers ToolHive's OAuth refresh.

Refactored `mapTransportError` into `mapStatusError` (status-driven) + `mapStringError` (string fallback) for readability and to stay under gocyclo 15.

Tests: `TestMapTransportError_5xxWithUnauthorizedBodyNotMisclassified` (500/502/503 cases), `TestMapCallError_5xxSurfacesUnchanged`.

### 2. Rehydrated transient Validate error inconsistency (MEDIUM — `transports.go:343`)

**Fixed.** `serveRehydrated` now distinguishes a transient error on a **cached** rehydrated session (→ 503, preserving the cache, matching the local path) from an **unknown** session that was never cached (→ 404 "Invalid session ID", unchanged). The distinction is `getRehydrated(sid) != nil`: if we have a cached reconstruction, the session WAS valid, so the error is likely transient.

Test: `TestRehydratedSession_TransientValidateError_Returns503` (flaky manager flakes once → 503, then session still cached and served on retry).

### 3. `extractResourceMetadataURL` stub (MINOR — `errors.go:77`)

**Fixed — no longer blocked by the go-sdk.** The shim's own `headerRoundTripper` has direct access to `resp.Header` and now captures `WWW-Authenticate` values into `errBody.wwwAuthHdrs` (in `captureErrorBody`). `extractResourceMetadataURL` reads from the captured header and parses the RFC 9728 §5.1 `resource_metadata` parameter using a direct port of mcp-go's well-tested `extractResourceMetadataURLs` parser (new file `rfc9728.go`). `AuthorizationRequiredError.ResourceMetadataURL` is now populated end-to-end through both `mapCallError` and `mapConnectError`.

Tests: `TestExtractResourceMetadataURL_FromCapturedHeader` (5 cases including quoted/token values, multi-param, empty), `TestMapCallError_Transport401_PopulatesResourceMetadataURL` (e2e), `TestMapConnectError_Transport401_PopulatesResourceMetadataURL` (e2e), `TestCaptureErrorBody_CapturesWWWAuthenticate` (e2e through the RoundTripper).

### 4. `normalizeObjectSchema` branch unification (NIT — `server.go:816`)

**Fixed.** Unified the two branches (empty-type-string and no-type-key) into a single `hasType && typeIsStr && typeStr == ""` / `!hasType` check, preserving the semantic distinction: empty-string type is mcp-go's "no schema declared" sentinel (always normalize), while an absent type without properties/required (e.g. `{$ref}`, `{oneOf}`, `{title}`) passes through verbatim. A non-string type value (e.g. `["object","null"]`) also passes through. All existing `TestNormalizeObjectSchema` cases still pass.

### 5. `notifCh` drain goroutine leak (pre-existing — `session.go`)

**Fixed.** `forgetSession` now closes `notifCh` via `sync.Once` (guarding against double-close when `sessionFor` re-creates an entry for the same ID). The drain goroutine in `newClientSession` exits when the channel is closed. A re-created session gets a fresh channel.

Test: `TestForgetSession_ClosesNotifChannel` (verifies close, double-forget safety, fresh channel on re-create).

### 6. `rehydrate` lock scope (LOW — `transports.go:443`)

**Documented** (per reviewer: "not worth restructuring now"). Added a scalability note to the `rehydrate` doc comment explaining the cross-sid serialization cost and the fix path if rehydration volume rises.

### General note (notifCh drain goroutine)

The drain goroutine leak you flagged in the general review note is fixed by item 5 above — `forgetSession` now closes `notifCh`, which lets the drain goroutine exit. The `sync.Once` guard addresses the double-close concern you raised.

mcpcompat/client/errors.go
JAORMX: Addressed in `eb17064`. The shim's own `headerRoundTripper` now captures `WWW-Authenticate` header values into `errBody.wwwAuthHdrs` (in `captureErrorBody`), and `extractResourceMetadataURL` parses the RFC 9728 §5.1 `resource_metadata` parameter from them (new file `rfc9728.go`, a direct port of mcp-go's well-tested parser). `AuthorizationRequiredError.ResourceMetadataURL` is now populated end-to-end through both `mapCallError` and `mapConnectError`. Tests: `TestExtractResourceMetadataURL_FromCapturedHeader` (5 cases), `TestMapCallError_Transport401_PopulatesResourceMetadataURL`, `TestMapConnectError_Transport401_PopulatesResourceMetadataURL`, `TestCaptureErrorBody_CapturesWWWAuthenticate` (e2e through the RoundTripper).

mcpcompat/client/errors.go
JAORMX: Addressed in `eb17064`. `mapTransportError` now surfaces a captured 5xx unchanged — the string fallback only runs when no status was captured at all (`h == nil || h.status == 0`). Implemented as a `mapStatusError` (returns `ok=true, err` for 5xx) + `mapStringError` split. Tests: `TestMapTransportError_5xxWithUnauthorizedBodyNotMisclassified` (500/502/503 with "unauthorized"/"401" in body), `TestMapCallError_5xxSurfacesUnchanged`.

mcpcompat/server/transports.go
JAORMX: Addressed in `eb17064`. `serveRehydrated` now distinguishes a transient error on a **cached** rehydrated session (→ 503, preserving the cache, matching the local path) from an **unknown** session that was never cached (→ 404 "Invalid session ID", unchanged). The distinction is `getRehydrated(sid) != nil`: if we have a cached reconstruction, the session WAS valid, so the error is likely transient. Test: `TestRehydratedSession_TransientValidateError_Returns503` (flaky manager flakes once → 503, then session still cached and served on retry).

mcpcompat/server/transports.go
JAORMX: Documented in `eb17064` — added a scalability note to the `rehydrate` doc comment explaining the cross-sid serialization cost and the fix path (build outside the lock, insert under it) if rehydration volume ever rises. Not restructuring now per your assessment.

mcpcompat/server/server.go
JAORMX: Addressed in `eb17064`. Unified the two branches into a single `hasType && typeIsStr && typeStr == ""` / `!hasType` check, preserving the semantic distinction: empty-string type is mcp-go's "no schema declared" sentinel (always normalize), while an absent type without properties/required (e.g. `{$ref}`, `{oneOf}`, `{title}`) passes through verbatim. A non-string type value (e.g. `["object","null"]`) also passes through. All existing `TestNormalizeObjectSchema` cases still pass.

mcpcompat/client/errors.go
jhrozek: Re-checked against `eb17064` (wave 6) — this one is **still open**. Wave 6's `mapStatusError`/`mapStringError` split fixed the *adjacent* case (a captured **5xx** whose body contains "unauthorized" is now surfaced unchanged), but that's a different code path from the one this comment is about, so it's easy to read the 5xx fix as covering both — it doesn't.

The case here is an **application-level JSON-RPC error returned at `initialize`** (200 OK + JSON-RPC error body ⇒ `h.status == 0`), which trace through the new code still misclassifies:

- `mapConnectError`: `h.status >= 400 && < 500` is false (status is 0) → `return mapTransportError(ctx, err, h)`
- `mapTransportError`: `h != nil && h.status >= 400` is false → skips `mapStatusError` → `return mapStringError(ctx, err)`
- `mapStringError`: matches `strings.Contains(msg, "401") || strings.Contains(msg, "unauthorized")` on the JSON-RPC error text → returns `ErrUnauthorized` + `AuthorizationRequiredError` ⇒ **false OAuth refresh**

`mapCallError` already guards exactly this (errors.go, the `if (h == nil || h.status == 0) && errors.As(err, &wireErr) { return err }` block); `mapConnectError` does not. `Connect` runs `initialize` through the same `handleSend` machinery as the call methods (go-sdk v1.6.1 client.go:271), so the same `*jsonrpc.Error` reaches `mapConnectError` with no captured status.

Suggested symmetric fix (`jsonrpc` is already imported in this file), added just before the final `return mapTransportError(...)` in `mapConnectError`:

```go
if h == nil || h.status == 0 {
	var wireErr *jsonrpc.Error
	if errors.As(err, &wireErr) {
		// RPC-level error with no captured HTTP status: an application-level
		// initialize rejection, not a transport failure. Surface unchanged
		// (mirrors mapCallError) rather than string-matching it into a
		// transport auth/legacy-SSE sentinel.
		return err
	}
}
```

Worth a `TestMapConnectError_DoesNotMisclassifyJSONRPCErrors` mirroring the existing `mapCallError` test. Scope caveat unchanged: this only bites when a server rejects `initialize` with an app-level JSON-RPC error whose text matches — narrower than the call path, but it's the exact class the guard was introduced to prevent.

JAORMX: Addressed in `c469b88`. Added the symmetric JSON-RPC guard to `mapConnectError`: when `h.status == 0` and `err` is a `*jsonrpc.Error`, the error is surfaced unchanged (mirrors `mapCallError`). An app-level JSON-RPC error at `initialize` (200 OK + error body) with "unauthorized"/"401" text is no longer misclassified as `ErrUnauthorized` → no false OAuth refresh. Test: `TestMapConnectError_DoesNotMisclassifyJSONRPCErrors` (3 cases: unauthorized-in-message, 401-in-message, method-not-allowed-in-message).

mcpcompat/client/client.go
JAORMX: Addressed in `c469b88` — went with the compat layer. Added `Client.SetLevel(ctx, mcp.SetLevelRequest) error` that forwards to `SetLoggingLevel(ctx, request.Params.Level)`, and re-exported `SetLevelRequest`/`SetLevelParams` from the mcp package. Downstream code using `c.SetLevel(ctx, mcp.SetLevelRequest{...})` now compiles against the shim unchanged. Test: `TestSetLevel_CompatAlias` (e2e against a live server with logging capabilities).

mcpcompat/client/client.go
JAORMX: Addressed in `c469b88`. Wired `opts.ResourceUpdatedHandler` (dispatches `notifications/resources/updated` with the `uri` param) and `opts.ElicitationCompleteHandler` (dispatches `notifications/elicitation/complete` with the `elicitationId` param), mirroring the existing progress/logging handlers. Updated the `OnNotification` doc comment to list all forwarded notification types. Test: `TestResourceUpdatedAndElicitationCompleteHandlers` (verifies both handlers are non-nil and dispatch the correct method strings + params).

mcpcompat/server/transports.go
JAORMX: Acknowledged and documented in `c469b88`. Strengthened the `WithHeartbeatInterval` doc to explain the passive keep-alive is load-bearing: elicitation and all server→client notifications (list_changed, progress, logging, resources/updated, elicitation/complete) structurally depend on the client's idle standalone SSE stream staying open, and any intermediary (LB, reverse proxy, ToolHive's own proxy) will reap that idle stream on timeout — after which elicitation fails (`ErrRejected`) and notifications are silently dropped. The TODO now says this should be prioritized before internet-facing vMCP use, and a max-session cap / idle sweep would also help. Tracked as `TODO(issue #156)`.

mcpcompat/server/transports.go
JAORMX: Documented in `c469b88`. Added a compat-note comment at the 503 site explaining: mcp-go returned 404 on any Validate error; the shim returns 503 on a non-terminated (transient) error and reserves 404 for genuine termination. 503 is NOT mapped by the client classifier (`errors.Is(err, ErrSessionTerminated)` is false), so a caller that keyed re-initialization off mcp-go's 404 → `ErrSessionTerminated` path won't re-init on a transient blip — it sees a generic retryable error. This is intentional (not dropping local state avoids the split-brain), but callers relying on that path should be aware.

JAORMX:
## Wave 7 — addresses jhrozek's second review (the one I missed)

My apologies — jhrozek posted a second set of 5 inline comments at `09:27:55Z` that I completely missed in wave 6. All 5 are now addressed in `c469b88`. `task lint` (0 issues), `task test` (-race, all green), `task license-check` (clean).

### 1. [BLOCKING] `mapConnectError` missing JSON-RPC guard (`errors.go:58`)

**Fixed.** `mapConnectError` now has the symmetric `*jsonrpc.Error` guard that `mapCallError` has: when `h.status == 0` and `err` is a `*jsonrpc.Error`, the error is surfaced unchanged. An app-level JSON-RPC error at `initialize` (200 OK + error body) with "unauthorized"/"401" text is no longer misclassified as `ErrUnauthorized` → no false OAuth refresh.

Test: `TestMapConnectError_DoesNotMisclassifyJSONRPCErrors` (3 cases).

### 2. [Question] `SetLevel` compat alias (`client.go:271`)

**Fixed — went with the compat layer.** Added `Client.SetLevel(ctx, mcp.SetLevelRequest) error` (forwards to `SetLoggingLevel`) and re-exported `SetLevelRequest`/`SetLevelParams` from the mcp package. Downstream code using `c.SetLevel(ctx, mcp.SetLevelRequest{...})` now compiles unchanged.

Test: `TestSetLevel_CompatAlias` (e2e).

### 3. [Compat gap] `OnNotification` drops `resources/updated` + `elicitation/complete` (`client.go:583`)

**Fixed.** Wired `opts.ResourceUpdatedHandler` (→ `notifications/resources/updated` with `uri` param) and `opts.ElicitationCompleteHandler` (→ `notifications/elicitation/complete` with `elicitationId` param). Updated `OnNotification` doc to list all 7 forwarded notification types.

Test: `TestResourceUpdatedAndElicitationCompleteHandlers`.

### 4. [Follow-up] No keep-alive (`transports.go:176`)

**Documented + acknowledged.** Strengthened `WithHeartbeatInterval` doc: the passive keep-alive is load-bearing (elicitation + all server→client notifications depend on the idle SSE stream), should be prioritized before internet-facing vMCP use. Tracked as `TODO(issue #156)`.

### 5. [Doc note] 503 diverges from mcp-go's 404 (`transports.go:358`)

**Documented.** Added a compat-note comment at the 503 site: 503 is not mapped by the client classifier (`errors.Is(err, ErrSessionTerminated)` is false), so callers keying re-init off mcp-go's 404→`ErrSessionTerminated` path won't re-init on a transient blip. Intentional (avoids split-brain), but callers should be aware.

jhrozek (approved):
Approving — all review findings are resolved. Verified against `c469b88` (build + client/mcp/server tests green, including the new `TestMapConnectError_DoesNotMisclassifyJSONRPCErrors`).

- **`mapConnectError` JSON-RPC guard** — fixed with the symmetric `errors.As(&wireErr)` guard + regression test.
- **`SetLevel` drop-in parity** — fixed via `Client.SetLevel(ctx, mcp.SetLevelRequest)` wrapper + `SetLevelRequest`/`SetLevelParams` aliases.
- **`notifCh` goroutine leak** — fixed (wave 6) with a `sync.Once` close in `forgetSession`.
- **`OnNotification` dropped notifications** — fixed by wiring `ResourceUpdatedHandler` and `ElicitationCompleteHandler`.
- **503 vs 404 divergence** — compat note added at the 503 site.
- **Keep-alive gap** — acknowledged and documented with a prioritized TODO (passive SSE keep-alive). Reasonable to land as a tracked follow-up; worth prioritizing before internet-facing vMCP use, alongside the session-reaping / max-session-cap work also tracked under issue #156.

Nice, thorough iteration across the waves. 🚀

PR #166, opened 2026-07-13
Add shared metrics helper package (meter, bucket presets, labels, lint)

glageju:
Closes https://github.com/stacklok/stacklok-enterprise-platform/issues/2298

## Summary

Adds a new `metrics` package to `toolhive-core` providing the shared histogram bucket presets and canonical (cross-component) label-key constants the ToolHive proxy, vMCP, and registry will standardize on, per the Metrics Standardization RFC (D6, §3.3). This is Phase 0 of the standardization effort — `toolhive-core` hosts the helper but emits no metrics of its own; downstream rename/coverage tasks import it.

The package does not include a meter-creation wrapper or a build-time naming/label lint (see Scope Changes below for why).

### `metrics` package (new)
- `BucketsFastHTTP()`, `BucketsMCPProxy()`, `BucketsLongRunning()` — histogram bucket boundary presets in seconds (0.005-10, 0.01-300, 0.1-300 respectively), each returning an independent slice.
- Canonical label-key constants (`LabelMCPServer`, `LabelOutcome`, `LabelMCPMethod`, `LabelToolName`, `LabelCompositeTool`, `LabelTransport`, `LabelErrorType`) as the single source of truth for label spellings. Scoped to canonical *common* keys only — concepts more than one component emits and a dashboard joins/groups on. Component-local keys (e.g. operator reconcile's `phase`, the circuit breaker's `from`/`to`) are intentionally not exported here.

### Dependencies
- No new dependencies. The package has no OTel import; consumers construct their own meters via `go.opentelemetry.io/otel` directly.

### Documentation
- Registers `metrics` in the package inventory tables in `README.md` and `CLAUDE.md` (Alpha stability).

## Scope Changes

The original implementation also included a `Meter(scope string) metric.Meter` wrapper over the global OTel meter provider, and two build-time lint functions (`ValidateName`, `ValidateLabelKind`, later renamed `ValidateLabel`) rejecting a `gateway` service segment, malformed dotted names, boolean-typed labels, and banned label re-spellings. Both were dropped before merge:

- **`Meter()`** was a pure pass-through to `otel.Meter(scope)` with no encoded decision (unlike `logging.New()`'s JSON/level/timestamp defaults) and no planned downstream consumer — the proxy/vMCP/registry rename tasks never reference it, and the LLM Gateway coverage-addition task constructs a brand-new meter via `otel.Meter(...)` directly rather than this wrapper.
- **The build-time lint** existed specifically to guarantee the canonical naming/label conventions hold for the LLM Gateway, the one emitter that cannot import this package (in-tree, `//go:build enterprise`). No downstream task ever wires `ValidateName`/`ValidateLabel` into a test, so it wasn't providing the enforcement it was designed for. The planned `.claude/rules/metrics.md` rule doc is the intended authority for that convention going forward.

The bucket presets and label constants stay: they're plain values a consumer imports instead of hand-typing, which needs no enforcement mechanism to be worth sharing.

## Implementation Details

- The bucket presets carry no package-level state; each call returns an independent slice, so mutating one caller's result cannot affect another.
- Metric namespace segments follow the RFC as amended: `ai_gateway` (renamed from `llm_gateway`) and the ToolHive proxy's `stacklok.toolhive.proxy.*` subsystem segment.
- Package shape (`doc.go` + implementation files + table-driven tests) mirrors the existing `logging/` package as the in-repo precedent for a shared utility package.

## Testing

- Unit tests cover each bucket preset's documented min/max boundaries and strict monotonicity, and independent slice mutation across repeated preset calls.
- `task lint`, `task license-check`, and `task test ./metrics/...` (including `-race`) pass; package coverage is 100%.


glageju (commented):
## Multi-Agent Consensus Review
**Agents consulted**: Go correctness specialist, test coverage specialist, API/architecture specialist, general code quality specialist, codex (gpt-5.5)

### Overall
This is a clean, well-scoped greenfield addition: a `metrics` package providing a `Meter()` wrapper, three histogram bucket-boundary presets, canonical label-key constants, and two build-time lint functions (`ValidateName`, `ValidateLabel`). The package mirrors the existing `logging/` package's shape, stays side-effect-free (verified by a dedicated `ManualReader`-backed test asserting zero `ScopeMetrics`), and its exported surface matches the linked issue's scope boundary exactly — no drift in either direction. Bucket presets correctly return a fresh slice per call rather than sharing package-level state, and the reflection-based `isBoolKind` helper correctly handles nil pointers, named bool types, and pointer chains. Test coverage is genuinely 100%, independently re-verified by the test-coverage reviewer via `go tool cover -func`, with every branch demonstrably hit by a distinct case.

Two review notes were addressed during the review: `ValidateName` now rejects segments containing whitespace, uppercase letters, or other characters outside `[a-z0-9_]`, matching the "well-formed dotted OTel metric name" guarantee the doc comment already claimed; and `ValidateLabelKind` is renamed to `ValidateLabel` since it runs two independent checks (banned key re-spelling, boolean-typed value) and the old name only described the latter.

A few remaining observations are informational only, each already weighed and accepted as a deliberate trade-off rather than a defect:
- `Meter()` returns `metric.Meter` directly, so every downstream importer's compile-time OTel version is tied to whatever `toolhive-core` pins. Inherent to building a thin OTel wrapper.
- `go.opentelemetry.io/otel/sdk/metric` lands as a direct `go.mod` dependency, used only in `metrics_test.go` — correct `go mod tidy` behavior given Go has no test-only dependency concept.
- Label keys are bare `string` constants rather than a distinct type, matching the constraint that OTel's `attribute.String(key string, ...)` only accepts plain strings.

### Documentation
No documentation gaps.

---
*Generated with Claude Code*

reyortiz3 (approved):

PR #168, opened 2026-07-16
Add pre-dispatch denial gate to Streamable HTTP

JAORMX:
## Summary

**Why:** The Streamable HTTP transport (go-sdk under the shim) writes every JSON-RPC response — success *or* error — at HTTP 200, and rejects an unknown tool with a hardcoded `-32602`. There is no hook for a handler or a host to signal an *authorization denial*. A host (e.g. ToolHive vMCP) that enforces policy above the SDK therefore has no way to return an explicit denial for a request it blocks: a denied call that was filtered out of the advertised tool set comes back as `-32602 tool not found` at HTTP 200, indistinguishable from a typo and invisible to audit. See stacklok/toolhive#5827.

**What:** Add `WithCallGate` — a generic, policy-free seam consulted once per POST, *before* session-ID validation and *before* SDK dispatch. On a non-nil `*Denial` it short-circuits and writes the host-chosen JSON-RPC error `code`/`message` at a host-chosen HTTP status (403 by default), reusing the shim's own `mcp.JSONRPCError` envelope so the denial is byte-identical in shape to every other error the shim emits. The request body is read only on the deny path (to echo the request id best-effort; `null` for an unparsable/batch/no-id body); the allow path carries zero overhead.

The shim carries **no policy** — it never interprets a `Denial` beyond its wire fields and reads only context the host populated. What "denied" means is entirely the host's concern.

## Design notes for reviewers

- **Placement** (in `StreamableHTTPServer.ServeHTTP`): after `contextFunc` (so a gate can read host-injected context), before the nonce bridge, before **both** the local go-sdk handler and the cross-replica rehydration path (so multi-replica deployments gate identically), and before session validation (403-before-404 — a denial is determinable without session state, matching hosts whose authz sits outside the SDK).
- **Body bounding** is deliberately left to the host: the read is unbounded here because the host is expected to cap the request body outermost (e.g. `http.MaxBytesReader`); a second limiter in this generic shim would redundantly hardcode a policy the host owns.
- **Scope:** Streamable HTTP only. The stdio bridge (`HandleMessage`) has no HTTP status and denies upstream; the legacy SSE transport is not wired. Both are documented as out of scope on the `CallGate` doc comment.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test plan

- [x] New unit tests in `mcpcompat/server/callgate_test.go` drive the real `*StreamableHTTPServer.ServeHTTP`: deny path (exact envelope: status, content-type, code, echoed id), allow path (dispatch runs, body untouched), no-gate regression, 403-before-404 (bogus + cross-replica session), unparsable/batch/empty/no-id body → `null` id, non-POST bypass, HTTPStatus default vs explicit, and a spy asserting dispatch is short-circuited on deny.
- [x] `task test` — pass
- [x] `task lint` — pass (0 issues)

### DIFF
--- a/mcpcompat/server/callgate.go
+++ b/mcpcompat/server/callgate.go
@@ -1,0 +1,145 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+
+	mcp "github.com/stacklok/toolhive-core/mcpcompat/mcp"
+)
+
+// CallGate is consulted once per HTTP POST on the Streamable HTTP transport,
+// BEFORE session-ID validation and BEFORE the message is dispatched to the
+// underlying SDK. The gate decides from the request context (the host's own
+// middleware has already run and populated it — identity, parsed request);
+// the shim passes the request through untouched.
+//
+// A nil return admits the request. A non-nil *Denial short-circuits dispatch:
+// the response is a JSON-RPC error envelope (the request's id echoed
+// best-effort; null when unparsable) with Denial.Code/Message, written at
+// Denial.HTTPStatus (403 when zero), Content-Type application/json.
+//
+// Ordering contract: the gate runs before session validation, so a denied
+// call with an invalid/terminated session ID receives the denial (e.g. 403),
+// not 404 — a denial is determinable without session state, and this matches
+// hosts whose authorization middleware sits outside the SDK entirely.
+//
+// Mechanism only: the shim carries no policy. It never inspects the meaning of
+// a Denial beyond its wire fields, and it reads context the host populated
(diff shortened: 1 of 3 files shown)


jhrozek (approved):
Reviewed the pre-dispatch denial gate. Clean, minimal, well-scoped seam.

- Zero-overhead allow path (body untouched unless denied), correct 403-before-404 ordering, and placement after `contextFunc` / before the nonce bridge are all correct and justified in-code.
- Tests are thorough (deny/allow/no-gate, 403-before-404 on both local and rehydration paths, id extraction for malformed/batch/empty/notification bodies, non-POST bypass, HTTPStatus default + override). Tests and lint pass locally.

Non-blocking notes for the consuming side:
- The gate is per-call (POST) only — GET/DELETE bypass it by design, so keep outer middleware as the primary authz boundary.
- `extractRequestID` reads the body unbounded on the deny path; confirm the host wraps requests with `http.MaxBytesReader` outermost.

✅ Approve.

PR #169, opened 2026-07-16
Implement passive SSE keep-alive on GET streams

JAORMX:
## Summary

**Why:** `WithHeartbeatInterval` has been an unwired no-op — it stored the interval but never emitted anything. An idle standalone GET SSE stream therefore produces **no periodic bytes**, so any intermediary (load balancer, reverse proxy, ToolHive's own proxy) reaps it on its idle timeout. That silently breaks **elicitation** and **every server→client notification** (`list_changed`, progress, logging, `resources/updated`, elicitation completion) — all of which structurally depend on the client's idle standalone SSE stream staying open — and it feeds unbounded session growth (abandoned sessions are never reaped). This is the passive keep-alive the `TODO(issue #156)` in `WithHeartbeatInterval` always intended.

**What:** Wire the interval to a passive keep-alive. `keepAliveWriter` wraps the GET-stream `http.ResponseWriter` and injects an SSE comment (`": keep-alive"`) at the configured interval, but only once the response is established as an SSE stream (status 200 + `Content-Type: text/event-stream`). It covers both the local go-sdk handler path and the cross-replica `serveRehydrated` path. POST (JSON) and DELETE responses are never wrapped.

## Why not go-sdk's `ServerOptions.KeepAlive`

Deliberately unused (and must stay so): go-sdk's `KeepAlive` sends an **active ping request** (server→client) each interval and `session.Close()`s on failure. Under `JSONResponse: true` mode that ping routes to the standalone SSE stream; a JSON-only client (no open GET stream — the shim client's default) has that stream unconnected, so the ping is rejected and **every healthy session is evicted on the first tick**. An SSE comment is invisible at the JSON-RPC layer and costs the client nothing. (This also diverges from mcp-go, which sent a full JSON-RPC ping *request* as its heartbeat — a conforming client would answer with a response carrying an ID the server never issued.)

## Concurrency & lifecycle

- All writes and flushes — the handler's `Write`/`WriteHeader`/`Flush` and the keep-alive tick — are serialized by a mutex. go-sdk writes one whole SSE frame per `Write`, so serializing at the call boundary preserves frame atomicity: a comment can never land mid-event.
- The per-connection ticker goroutine is started only for an established SSE 200 (error responses — the denial gate's 403, session 404/409/503 — never spawn one) and is torn down by a `defer`red, idempotent `stopKeepAlive` on every `ServeHTTP` return (including panic unwind). No goroutine leak; no write after `ServeHTTP` returns.

## Also

Drops the dead `time.Duration` keep-alive parameter that was threaded through `buildServer`/`getServerFunc` and ignored at all 7 call sites (a silent no-op parameter).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test plan

- [x] New `keepalive_internal_test.go` (wrapper): comments emitted at ~interval each flushed; implicit-header/replay path; ticker **never** starts on non-SSE / non-200 / SSE-headers-with-no-body / stopped-before-start; `Content-Type` with parameters (`; charset=utf-8`) still starts; first comment is not immediate (one full interval); `stopKeepAlive` prevents later writes and the goroutine exits; write-error exits; a `-race`-meaningful concurrency test hammering `Write` with whole frames alongside the ticker asserts no interleaving.
- [x] `keepalive_test.go` (end-to-end): GET stream receives comment bytes; the initialize/POST response body contains **no** keep-alive bytes; heartbeat disabled → idle GET stays silent; rehydrated GET path covered.
- [x] `task test` (with `-race`) — pass; 15× race loop of the keep-alive tests — no flakes.
- [x] `task lint` — pass (0 issues).

### DIFF
--- a/mcpcompat/server/keepalive.go
+++ b/mcpcompat/server/keepalive.go
@@ -1,0 +1,193 @@
+// SPDX-FileCopyrightText: Copyright 2026 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"log/slog"
+	"mime"
+	"net/http"
+	"sync"
+	"time"
+)
+
+// keepAliveComment is the SSE comment written on each heartbeat tick. SSE
+// comments (lines beginning with ':') are ignored by every spec-conforming SSE
+// parser: go-sdk's client-side scanEvents cuts each line on the first ':', and
+// an empty key yields no event. The payload MUST contain the colon — a line
+// without one is a malformed SSE event. This deliberately diverges from
+// mcp-go, which sent a full JSON-RPC ping REQUEST as the heartbeat: a
+// conforming client would answer that ping with a response carrying an ID the
+// go-sdk server never issued, generating unknown-request-ID handling on every
+// tick. A comment is invisible at the JSON-RPC layer and costs the client
+// nothing.
+var keepAliveComment = []byte(": keep-alive\n\n")
+
+// keepAliveWriter wraps the http.ResponseWriter handed to the go-sdk GET handler
+// and injects keepAliveComment at each heartbeat interval once — and only once —
+// the response is established as an SSE stream (status 200 and Content-Type
+// text/event-stream). All writes and flushes, the handler's and the keep-alive
+// tick's alike, are serialized by mu. go-sdk writes one whole SSE frame per
+// Write call (writeEvent buffers the full frame before issuing a single Write),
+// so serializing at the call boundary preserves frame atomicity: a comment can
+// never land in the middle of an event. This depends on go-sdk v1.6.1's
+// frame-atomic writeEvent; the module is version-pinned and the end-to-end test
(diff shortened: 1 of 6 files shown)


jhrozek (approved):
Reviewed the full diff (traced all 6 files) with an independent adversarial concurrency pass.

- `go vet` clean, `go test -race` green, all `buildServer`/`getServerFunc` callers updated, go-sdk pinned at v1.6.1 (matches the frame-atomicity assumption).
- Teardown is provably race-free without joining the ticker goroutine; denial/non-SSE/non-200 responses correctly never start the ticker; both local and rehydrated GET paths covered by one wrap point.
- Excellent test coverage (branch-complete unit tests + race-meaningful concurrency stress + e2e).

Only INFO-level notes, none blocking: the no-interleaving guarantee rests on go-sdk v1.6.1's whole-frame-per-Write behavior (documented + version-pinned, but not tested against go-sdk's actual write granularity) — worth a tripwire on the next dependency bump.

LGTM 🚀

PR #171, opened 2026-07-17
Fix mcpcompat dropping results that embed a resource

### LINKED ISSUE #170: mcpcompat drops entire tool/prompt result when it contains an embedded resource
## Summary

`mcpcompat.EmbeddedResource` has a `Resource ResourceContents` field where `ResourceContents` is a **non-empty interface**, and `EmbeddedResource` has **no custom `UnmarshalJSON`**. So `json.Unmarshal` of any `"type":"resource"` content item errors:

```
json: cannot unmarshal object into Go struct field EmbeddedResource.resource of type mcp.ResourceContents
```

Because `CallToolResult.UnmarshalJSON` (and the client's `convertResult`) propagate that error, **any tool result — or prompt message — containing an embedded resource is lost in its entirety**, including any sibling text/image parts in the same content array.

## Impact

Discovered from a ToolHive vMCP MCP-conformance run (the mcp-go → mcpcompat swap). Two conformance scenarios regressed:

- `tools-call-embedded-resource` → "No resource content found"
- `tools-call-mixed-content` → "Expected multiple content items; Missing image content; Missing resource content"

The `prompts/get` path is affected too, since `convertGetPromptResult` also decodes content via `mcp.UnmarshalContent`.

## Call chain

1. `client.CallTool` converts the go-sdk result via the generic `convertResult` → `jsonConvert`.
2. That runs `CallToolResult.UnmarshalJSON`, which calls `UnmarshalContent` per item.
3. `UnmarshalContent`, case `ContentTypeResource`, does `json.Unmarshal(data, &EmbeddedResource{})` — stdlib cannot populate the `ResourceContents` interface field → error.
4. The error aborts the whole decode; `convertResult` discards the partially-decoded result.

`ReadResource` and `GetPrompt` were hand-written to avoid this exact trap (`convertReadResourceResult` maps concrete text/blob types explicitly); `CallTool` was left on the generic path.

## Fix

Add a custom `UnmarshalJSON` to `EmbeddedResource` that resolves the polymorphic `Resource` field to a concrete `TextResourceContents`/`BlobResourceContents` (blob-precedence, mirroring `convertReadResourceResult`). `UnmarshalContent` is the single chokepoint backing both `tools/call` and `prompts/get`.

## Repro

```go
wire := []byte(`{"type":"resource","resource":{"uri":"file:///x.txt","mimeType":"text/plain","text":"body"}}`)
_, err := mcp.UnmarshalContent(wire)
// err: json: cannot unmarshal object into Go struct field EmbeddedResource.resource of type mcp.ResourceContents
```

jhrozek:
## What

Adds a custom `UnmarshalJSON` to `mcpcompat.EmbeddedResource` so a `"type":"resource"` content item decodes correctly instead of erroring.

## Why

`EmbeddedResource.Resource` is the `ResourceContents` **interface** with no custom unmarshaler, so `json.Unmarshal` of any embedded-resource content item failed:

```
json: cannot unmarshal object into Go struct field EmbeddedResource.resource of type mcp.ResourceContents
```

That error propagated out of `CallToolResult.UnmarshalJSON` / the client's `convertResult`, discarding the **entire** result — so a tool result (or prompt message) containing an embedded resource lost all of its content, including sibling text/image parts.

`ReadResource`/`GetPrompt` were hand-written to sidestep this; `CallTool` was on the generic round-trip path. The new `UnmarshalJSON` lives at `UnmarshalContent`, the single chokepoint backing both `tools/call` and `prompts/get`, so one method fixes both.

## How

Resolve the polymorphic `Resource` field to a concrete `TextResourceContents`/`BlobResourceContents`, using blob-precedence to mirror the client's existing `convertReadResourceResult`.

## Tests

- `UnmarshalContent` decodes text and blob embedded resources to the right concrete type.
- `CallToolResult` mixed content `[text, image, resource]` — all three survive.
- Round-trip fidelity (decode → marshal → decode).

`task` passes (test + lint + vet), license headers verified.

Fixes #170

### DIFF
--- a/mcpcompat/mcp/content.go
+++ b/mcpcompat/mcp/content.go
@@ -122,6 +122,56 @@ type EmbeddedResource struct {
 
 func (EmbeddedResource) isContent() {}
 
+// embeddedResourceJSON is a helper for unmarshaling EmbeddedResource, whose
+// Resource field is the ResourceContents interface and cannot be populated by
+// a generic JSON round-trip.
+type embeddedResourceJSON struct {
+	Annotated
+	Meta     *Meta           `json:"_meta,omitempty"`
+	Type     string          `json:"type"`
+	Resource json.RawMessage `json:"resource"`
+}
+
+// UnmarshalJSON decodes an embedded resource, resolving the polymorphic
+// Resource field to a concrete TextResourceContents or BlobResourceContents.
+// Without this, stdlib cannot unmarshal into the ResourceContents interface and
+// any "type":"resource" content item errors, aborting the whole result decode.
+func (e *EmbeddedResource) UnmarshalJSON(data []byte) error {
+	var raw embeddedResourceJSON
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	e.Annotated = raw.Annotated
+	e.Meta = raw.Meta
+	e.Type = raw.Type
+
+	if len(raw.Resource) == 0 {
+		return nil
+	}
+	// Peek to decide text vs blob. A non-empty "blob" field means blob contents;
+	// otherwise treat as text (mirrors the client's convertReadResourceResult).
+	var probe struct {
+		Blob string `json:"blob"`
+	}
+	if err := json.Unmarshal(raw.Resource, &probe); err != nil {
+		return err
+	}
+	if probe.Blob != "" {
+		var blob BlobResourceContents
+		if err := json.Unmarshal(raw.Resource, &blob); err != nil {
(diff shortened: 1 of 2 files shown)


jhrozek:
Closing in favor of #173, which fixes the same bug (#170). #173 puts the fix in a dedicated `EmbeddedResource.UnmarshalJSON` method rather than inline in `UnmarshalContent` — the more idiomatic placement, and it also carries the v0.0.31 release note.

PR #173, opened 2026-07-17
Fix EmbeddedResource JSON unmarshaling

JAORMX:
## Summary

- **Why:** `EmbeddedResource.Resource` is an interface-typed field (`ResourceContents`) with no custom `UnmarshalJSON`. `UnmarshalContent`'s `json.Unmarshal` into a plain `EmbeddedResource` therefore errored on **every** embedded-resource content item — `cannot unmarshal object into Go struct field EmbeddedResource.resource of type mcp.ResourceContents`. That aborted the entire `CallToolResult` / `GetPromptResult` decode, so a downstream consumer (ToolHive vMCP) dropped the **whole** result — including sibling text/image content — whenever a tool or prompt returned an embedded resource. Verified end-to-end: `client.CallTool` against a live go-sdk server hard-errors on the resource item.
- **What:** add `EmbeddedResource.UnmarshalJSON`, decoding the `resource` object into `TextResourceContents` or `BlobResourceContents` (by presence of `blob` vs `text`), mirroring the existing interface-decoding shims for `ReadResourceResult` and `PromptMessage` content. Fixes both the tool-call and prompt embedded-resource paths in one place.

Found by pointing the MCP conformance suite at a vMCP endpoint (`tools-call-embedded-resource` and `tools-call-mixed-content` failed).

## Test plan

- [x] `TestUnmarshalContent_EmbeddedResource_Text` / `_Blob` / `_Absent` — the resource object decodes into the right concrete type with fields intact (URI/MIMEType/Text or Blob), and the empty case doesn't panic.
- [x] `TestCallToolResult_Unmarshal_WithEmbeddedResource_And_Siblings` — a result with `[text, image, embedded-resource]` round-trips with **all three** items surviving (the exact downstream mixed-content failure).
- [x] Mutation-verified: without the new `UnmarshalJSON`, the siblings test fails; with it, passes.
- [x] `go build ./...`, `task lint` (0 issues), `task test` (full suite) green.

## Note for reviewer

Needs a **release (v0.0.31)** once merged — the ToolHive side just bumps the dependency (its own conversion is already correct; the content simply never reached it). This also fixes embedded resources in `GetPrompt`, not just tool calls.

### DIFF
--- a/mcpcompat/mcp/content.go
+++ b/mcpcompat/mcp/content.go
@@ -122,6 +122,70 @@ type EmbeddedResource struct {
 
 func (EmbeddedResource) isContent() {}
 
+// embeddedResourceJSON is a helper type for unmarshaling EmbeddedResource. It
+// mirrors EmbeddedResource but keeps the resource as a raw message so the
+// ResourceContents interface field can be decoded into a concrete type.
+type embeddedResourceJSON struct {
+	Annotated
+	Meta     *Meta           `json:"_meta,omitempty"`
+	Type     string          `json:"type"`
+	Resource json.RawMessage `json:"resource"`
+}
+
+// UnmarshalJSON implements custom JSON unmarshaling for EmbeddedResource.
+// The Resource field is a ResourceContents interface, which encoding/json
+// cannot populate directly. The nested resource object is decoded into the
+// concrete TextResourceContents or BlobResourceContents type based on the
+// presence of the "blob" (blob resource) or "text" (text resource) field,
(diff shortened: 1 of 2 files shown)


jhrozek (approved):
LGTM. Fixes #170 cleanly: follows the existing `toolResultContentJSON` helper pattern in the same file, mirrors all EmbeddedResource fields, and handles null/absent gracefully. Tests reproduce the real mixed-content downstream failure. Tests + vet green locally.

PR #187, opened 2026-07-24
Serve MCP 2026-07-28 stateless over Streamable HTTP

JAORMX:
## Summary

**Why:** PR1 (#185) put go-sdk v1.7.0-pre.3 in place. This is **PR2 of 3** — the actual stateless *serving* capability. It lets `mcpcompat`'s Streamable HTTP server answer the MCP **2026-07-28** revision (POST-only, `GET`/`DELETE` → 405, a fresh temporary session per POST with no `Mcp-Session-Id`, `server/discover` answered natively) via go-sdk's `StreamableHTTPOptions.Stateless`. It's the first piece that will let downstream ToolHive retire its hand-rolled Modern dispatch.

**What:** a new additive `WithStateless(bool)` option, plus the work to make the shim's per-session projection machinery correct **and safe** under stateless — where there's no `initialize` handshake and no `Mcp-Session-Id`.

## The security-critical part: per-request isolation

Under stateless, go-sdk hands each POST a temporary session with `ID() == ""`. The shim must **never** key its shared session registry on the empty string — otherwise every concurrent request would share one `clientSession` (one `boundServer`, one tools map) and leak one identity's projected tools to another.

- Each stateless dispatch gets its own **unstored ephemeral `clientSession`**, bound to that request's **per-request go-sdk server** (`getServer` runs per POST), so concurrent identities are fully isolated at both the shim and SDK layers.
- The ephemeral-vs-shared decision is gated on the server's **stateless mode**, not merely `ID() == ""`. This closes the `allowsessionsinstateless` compat-flag path (under which a stateless request *can* carry a client-supplied session id) — it can never route back to the shared registry. `registerAndSync` is likewise gated off under stateless.
- **Isolation test** (`-race`, looped): 50 concurrent goroutines across two identities, each asserting it sees only its own projected tool — plus a variant where every request carries the **same stale `Mcp-Session-Id`**, proving the guard holds even in the compat-flag scenario. Verified to fail if the shared-registry bug is reintroduced.

## Projection re-anchored onto the before-hooks

`OnRegisterSession` never fires statelessly (no `initialize`), so per-identity projection runs via `fireBeforeHooks` per request instead.

⚠️ **Known limitation (documented on `WithStateless`):** this covers **tools only**. Resources, resource-templates, and prompts fall back to the server-level set, because the drop-in Hooks API (mirroring `mark3labs/mcp-go`) has no before-hook seam for them. Expanding that API is intentionally out of scope for this PR; it's a tracked follow-up for when a consumer needs per-identity resource/prompt projection under stateless. (No consumer wires this shim path for stateless projection yet.)

## Other correctness

- `SessionIdManager` is **inert** under stateless (`Generate`/`Validate`/`Terminate` never consulted; the generator isn't even installed) — asserted by test.
- The ephemeral session's notification channel is closed at end of request (once-guarded), so its drain goroutine can't leak — even on handler panic (deferred cleanup).
- `ServeHTTP` short-circuits to the go-sdk handler **after** the identity contextFunc + authorization/denial gate + nonce bridge, so stateless POSTs are still authorized and identity-scoped, and a stale client-sent `Mcp-Session-Id` can't reach rehydration/validation/DELETE.
- Legacy (<2026-07-28) clients are still served under stateless, in a degraded sessionless form — documented; it's not Modern-only.

## Deferred (tracked, commented in-code)
go-sdk stamps `cacheScope:"public"` on 2026-07-28 `list`/`discover` **result bodies**. It's inert while `ttlMs == 0` (spec: zero TTL ⇒ immediately stale), but a per-identity-projected result should carry `cacheScope:"private"` via an **in-body rewrite** (here or in the vMCP envelope — *not* an HTTP `Cache-Control` header, which an MCP-aware cache wouldn't honor) once a positive TTL is in play. Deferred; data itself is correctly isolated today.

## Reviewer notes
- The isolation test is a probabilistic concurrency guard (50×100 contended ops under `-race`); it fails deterministically enough when the bug is present.
- Every stateless POST rebuilds a go-sdk server and spins a short-lived drain goroutine — heavier than stateful's once-per-session, acceptable.

## Depends on
Builds on #185 (already merged). Same **pre-release** constraint applies — the `release.yml` guard blocks tagging a release until go-sdk v1.7.0 final.

## Test plan
- [x] `task test` (race) — all suites green incl. the looped isolation tests (ran ×3 extra for flake confidence)
- [x] `task lint` — 0 issues

### DIFF
--- a/mcpcompat/server/server.go
+++ b/mcpcompat/server/server.go
@@ -470,7 +470,14 @@ func (s *MCPServer) AddPrompt(prompt mcp.Prompt, handler PromptHandlerFunc) {
 // syncs that session's overlay tools, resources and prompts onto its own server
 // once the OnRegisterSession hooks have run. This mirrors mcp-go, whose
 // per-session tools were dispatched per connection.
-func (s *MCPServer) buildServer(genSessionID func() string) (*gosdk.Server, error) {
+//
+// stateless MUST be true only when this server backs a StreamableHTTPServer
+// configured WithStateless(true); it is threaded through to
+// sessionDispatchMiddleware, which forces the per-request ephemeral session
+// binding for every dispatch in that mode (see bindSessionForDispatch) rather
+// than trusting the go-sdk ServerSession's ID(). Every other call site
+// (ServeStdio, SSEServer, StreamableHTTPServer.rehydrate) passes false.
+func (s *MCPServer) buildServer(genSessionID func() string, stateless bool) (*gosdk.Server, error) {
 	s.mu.RLock()
 	tools := make(map[string]ServerTool, len(s.tools))
 	for k, v := range s.tools {
@@ -549,7 +556,7 @@ func (s *MCPServer) buildServer(genSessionID func() string) (*gosdk.Server, erro
 		return nil, err
 	}
 
-	srv.AddReceivingMiddleware(s.sessionDispatchMiddleware(srv))
+	srv.AddReceivingMiddleware(s.sessionDispatchMiddleware(srv, stateless))
 	return srv, nil
 }
 
@@ -702,11 +709,12 @@ func addGlobalTool(srv *gosdk.Server, gt *gosdk.Tool, h gosdk.ToolHandler, name 
 
 // getServerFunc returns a getServer callback for the go-sdk HTTP/SSE handlers,
 // which invoke it once per new client session. genSessionID (may be nil) is the
-// session-ID generator to install on each per-session server. On a build error
-// it logs and returns nil, which the go-sdk handler surfaces as an HTTP 400.
-func (s *MCPServer) getServerFunc(genSessionID func() string) func(*http.Request) *gosdk.Server {
+// session-ID generator to install on each per-session server. stateless is
+// forwarded to buildServer (see its doc comment). On a build error it logs
+// and returns nil, which the go-sdk handler surfaces as an HTTP 400.
+func (s *MCPServer) getServerFunc(genSessionID func() string, stateless bool) func(*http.Request) *gosdk.Server {
 	return func(*http.Request) *gosdk.Server {
-		srv, err := s.buildServer(genSessionID)
+		srv, err := s.buildServer(genSessionID, stateless)
 		if err != nil {
 			if s.logger != nil {
 				s.logger.Error("building per-session MCP server", "error", err)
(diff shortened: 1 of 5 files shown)


jhrozek (approved):
## ✅ Approve

Reviewed `origin/main...HEAD` (commit `5359328`): stateless MCP 2026-07-28 Streamable HTTP serving for the mcp-go compat shim. Verified locally with `go test -race`, `go vet`, and `golangci-lint` — all clean.

### Isolation invariant — correctly enforced
The central risk is concurrent stateless POSTs collapsing onto a shared `sessionFor("")` entry (cross-identity tool/resource bleed). This is closed on three layers:
- `bindSessionForDispatch` forces the per-request ephemeral path on `stateless || ss.ID()==""`;
- `contextWithSession` / `sessionFor` / `registerAndSync` / `forgetSession` all guard `id==""`;
- go-sdk calls `getServer(req)` per POST, so each stateless request gets a fresh `*gosdk.Server`, isolating the tool overlay.

The `stateless` gate is genuinely **not** redundant with `ss.ID()==""` — the `allowsessionsinstateless=1` MCPGODEBUG path is a real bypass, and there's a direct unit regression for it plus two concurrent race tests (empty-ID and shared-attacker-ID, 50×100) confirming no bleed.

### Also verified
- Auth/denial gate runs **before** the stateless short-circuit (403-before-404 preserved).
- Ephemeral `clientSession` cleanup closes the notif channel symmetrically with `forgetSession`, so the per-request drain goroutine can't leak.
- `SessionIdManager` inertness under stateless is asserted; the stateful path has an explicit regression guard.

### Non-blocking notes
- **Perf (INFO):** every stateless POST rebuilds a full go-sdk server (map copy) + spawns a drain goroutine. Inherent to go-sdk's getServer-per-request contract, not a regression — just relevant when sizing high-QPS stateless deployments.
- **Known limitation (LOW):** only the tools before-hooks survive statelessly; resources/templates/prompts come from the global set (documented; a scoping gap, not a leak). Tracked follow-up.
- **cacheScope TODO (INFO):** deferral is sound — `ttlMs=0` is immediately-stale so nothing caches it yet; fix only needed once a positive TTL appears. Consider moving the in-path prose to the tracking issue.

Code comments are unusually thorough and accurate against `go-sdk@v1.7.0-pre.3`. Nice work.

PR #192, opened 2026-07-27
Export Sigstore bundle retrieval, offline verification, and identity API

### LINKED ISSUE #176: Export Sigstore bundle retrieval, offline verification, and identity extraction from container/verifier
## Summary

[RFC THV-0080](https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0080-skills-lock-file.md) (project-level skills lock file, merged) needs to extend Sigstore signature verification to skill OCI artifacts and git commits, reusing this package's existing `sigstore-go` integration (currently used for MCP server image verification) rather than building a second, independent implementation.

The toolhive `pkg/skills/verifier` package (Stack 2 of the RFC's implementation, building on the lock-file stack currently in review — see [stacklok/toolhive#5894](https://github.com/stacklok/toolhive/pull/5894) onward) is designed as a **thin wrapper** over `container/verifier`, per the RFC's own design section:

> Extend toolhive-core's existing `container/verifier` package (`sigstore-go`, already used for MCP image verification in `thv run --image-verification`) for skill OCI referrers.

Today, `container/verifier`'s bundle-retrieval, offline-verification, and identity-extraction internals (`bundleFromSigstoreSignedImage`, `bundleFromAttestation`, `sigstoreBundle`, `signerIdentityFromCertificate`, `compareBaseProperties` in `sigstore.go`/`attestations.go`/`verifier.go`) are all unexported. `Sigstore.VerifyServer` only supports the MCP-image online-verification flow; there's no way for a caller to fetch a bundle for storage or re-verify one offline against the embedded TUF trust roots.

## Requested API

Three additions to `container/verifier`, keeping the same package (not a new one) since they operate on the same `sigstoreBundle`/TUF-trust-root machinery already there:

1. **Bundle retrieval** — for storing a verified bundle alongside a skill's local install state, so `sync --check` can re-verify offline later without re-pulling from the registry:
   ```go
   // RetrieveBundles fetches every Sigstore bundle attached to imageRef (as a
   // cosign signature-manifest layer or an OCI referrer/attestation), covering
   // both discovery mechanisms container/verifier already understands.
   func RetrieveBundles(ctx context.Context, imageRef string, keychain authn.Keychain) ([]Bundle, error)

   // Bundle is a retrieved Sigstore bundle: the parsed form for immediate use,
   // and canonical serialized bytes (protojson) for durable storage.
   type Bundle struct {
       Parsed *bundle.Bundle
       Raw    []byte
   }
   ```

2. **Offline verification** — re-verify a stored bundle against the embedded TUF trust roots without contacting Rekor, for a hermetic CI check:
   ```go
   // VerifyBundleOffline verifies bundleBytes (as produced by RetrieveBundles)
   // against artifactDigest using only the embedded TUF trust roots — no
   // network access, suitable for CI.
   func (s *Sigstore) VerifyBundleOffline(bundleBytes []byte, artifactDigest string) (*verify.VerificationResult, error)
   ```

3. **Identity extraction** — recover the publisher identity from a verification result, for TOFU-pinning in a lock file:
   ```go
   // Identity is the Sigstore-verified publisher identity extracted from a
   // Fulcio certificate.
   type Identity struct {
       SignerIdentity      string
       CertIssuer          string
       SourceRepositoryURI string
   }

   // IdentityFromResult extracts Identity from a completed verification result.
   func IdentityFromResult(result *verify.VerificationResult) (Identity, error)
   ```

## Scope notes

- **Gitsign (git commit signature) verification is explicitly out of scope for this issue** — it's git-specific with no container/OCI coupling, so it will live directly in toolhive's `pkg/skills/verifier` rather than here.
- **Catalog-supplied expected identity** (matching a registry-declared publisher identity at first install, vs. today's trust-on-first-use) is also out of scope — tracked as separate follow-up work once `registry/types.Skill` (or similar) gains a `Provenance` field.
- This is a **blocking prerequisite** for the toolhive-side `pkg/skills/verifier` PR (first PR of RFC THV-0080's Stack 2, filed after Stack 1 — currently [#5892](https://github.com/stacklok/toolhive/pull/5892) through [#5897](https://github.com/stacklok/toolhive/pull/5897) — is fully merged). A toolhive-core release + `go.mod` bump will be needed before that PR can land.

## References

- RFC: https://github.com/stacklok/toolhive-rfcs/blob/main/rfcs/THV-0080-skills-lock-file.md
- POC (superseded by the stack above): https://github.com/stacklok/toolhive/pull/5715
- Existing verifier: `container/verifier/{verifier,sigstore,attestations,utils}.go`

samuv:
## Summary

ToolHive's skills lock-file feature (RFC THV-0080, Stack 2 — tracked in stacklok/toolhive#5899) needs to verify skill artifacts on install and re-verify stored bundles during `sync --check`, without duplicating this package's Sigstore internals. This promotes the existing machinery in `container/verifier` to a public API:

- **`RetrieveBundles(imageRef, keychain) ([]Bundle, error)`** — surfaces the bundles from both layouts the package already handles internally (cosign `.sig`-tag signature manifests and attestations). Each `Bundle` carries its canonical JSON serialization (`Raw`) so consumers can store it durably and re-verify later. Returns `ErrNoBundles` for unsigned artifacts.
- **`VerifyBundle(b, trustedMaterial, expected *Identity, opts...)`** — when `expected` is non-nil, the signer identity is **bound into the Sigstore verification policy itself** (`verify.WithCertificateIdentity`) rather than compared post-hoc in Go; `nil` expected is the trust-on-first-use case (chain-of-trust only, caller records the identity from the result). GitHub-Actions-derived identities match the SAN via a repo-anchored prefix.
- **`VerifyBundleOffline(raw, digestAlgo, digestHex, expected)`** — genuinely offline re-verification of a stored bundle against an **embedded** public-good `trusted_root.json` (new `OfflineTrustedMaterial()`): no TUF refresh, no network. The embedded root is a point-in-time snapshot, documented as such; live-freshness callers keep using `New`'s TUF fetch.
- **`VerifyBundleWithKey(b, pubKeyPEM)`** — the cosign key-pair flow: key-signed bundles carry no certificate or transparency material, so this uses `PublicKeyMaterial` + `WithNoObserverTimestamps` + a `WithKey` policy, kept as an explicit entry point rather than auto-detection.
- **`Identity` / `IdentityFromResult`** — exports the signer-identity extraction (including the GitHub Actions issuer normalization in `signerIdentityFromCertificate`).

Existing exported surface (`New`, `VerifyServer`, `GetVerificationResults`) is unchanged. `github.com/sigstore/sigstore` moves from indirect to direct (cryptoutils/signature used by the key-material path).

Fixes #176

## Test plan

- Key-signed bundle round trip: sign with an ephemeral key via sigstore-go → verify through `VerifyBundleWithKey` (this is the contract ToolHive's signing E2E will rely on); wrong-digest and wrong-key rejections
- `OfflineTrustedMaterial` parses the embedded root and carries Fulcio CAs + Rekor logs
- `VerifyBundleOffline` round-trips stored `Raw` bundles (failure at verification, not parsing) and rejects malformed input
- Identity policy construction for plain SAN, GitHub-Actions repo-anchored, nil (TOFU), and empty (rejected) identities; an expected identity fails closed against a certificate-less bundle
- `go test ./container/...` and `task lint` green


JAORMX:
Multi-axis panel review, **verified against sigstore-go v1.2.2 source** (module cache) and its docs. Posting findings; happy to discuss any.

## ✅ Verified correct against the sigstore-go API

These were checked directly in `sigstore-go@v1.2.2` source, not assumed:

- **Identity binding is genuinely in-policy, not post-hoc.** `WithCertificateIdentity` appends to `policy.certificateIdentities` (`pkg/verify/signed_entity.go:433`), and `Verify()` calls `policy.certificateIdentities.Verify(certSummary)` **inside** the flow (`:802`), setting `result.VerifiedIdentity`. The GitHub-Actions SAN regex is anchored both ends, case-sensitive, and `QuoteMeta`'d — no prefix/substring bypass.
- **`WithKey` ⇔ `WithCertificateIdentity` are mutually exclusive** (`:438`,`:452`), and `Verify()` rejects cert-bearing bundles under `WithKey` ("expected key signature, not certificate", `:632`). Key-pair path can't be confused with the cert path.
- **Offline verification still verifies the tlog cryptographically** — `VerifyTlogEntry` (`pkg/verify/tlog.go:48`) verifies the SET + inclusion proof against the embedded Rekor keys, no network, and enforces `MaxAllowedTlogEntries` / rejects duplicate entries. Offline only skips *freshness* (live STH), matching upstream's offline model.
- **The embedded `trusted_root.json` is justified, not reinvention.** sigstore-go embeds only the TUF *bootstrap* `root.json` (to start a refresh), **not** the `trusted_root.json` *target*; its own verification example (`examples/sigstore-go-verification/main.go:167`) loads exactly this file via `root.NewTrustedRootFromJSON`. Vendoring the target is the canonical offline pattern.
- **Backward compat:** existing exported surface (`New`/`VerifyServer`/`GetVerificationResults`) is byte-identical; the only `sigstore.go` change is `"sha256"`→`DigestAlgorithmSHA256` (improvement).
- **Duplication:** lean — correctly reuses `getSigstoreBundles`/`signerIdentityFromCertificate`/`verifierOptions`; Rule-of-Three not triggered on the Verify* skeletons (two genuinely different trust models).

## 🔴 Ship-blocker

**Panic on attacker-controlled OCI annotations — now on a public trust boundary.** `container/verifier/sigstore.go:218,223,231,256-257` — `getVerificationMaterialTlogEntries` uses single-value type assertions on the registry-supplied `dev.sigstore.cosign/bundle` annotation (`jsonData["Payload"].(map[string]interface{})`, `jsonData["apiVersion"].(string)`, etc.). A malformed annotation **panics** rather than errors. Pre-existing, but `RetrieveBundles` (`bundles.go:83`) now exposes it publicly → a registry/manifest attacker can crash the verifier process (DoS, CWE-248). Fix: two-value `ok` assertions returning the existing error path.

## 🟠 Spec mismatch (vs #176)

The PR ships a superset API whose shape doesn't match the three requested signatures, so the toolhive-side thin wrapper can't be written against the spec:

1. `RetrieveBundles` drops the spec'd `ctx context.Context` param (`bundles.go:83`) — no caller-side cancellation of registry fetches.
2. `VerifyBundleOffline` is a package-level func with split `digestAlgo/digestHex` + an added `expected` param, not the spec'd `(s *Sigstore) VerifyBundleOffline(bundleBytes, artifactDigest)`.
3. `IdentityFromResult` returns `*Identity` (spec: `Identity` value).
4. **Out-of-scope surface:** `VerifyBundle` implements the "match a registry-declared publisher at first install" semantics #176 lists **verbatim as OUT of scope**; `VerifyBundleWithKey` + `PublicKeyMaterial` (key-pair flow) aren't requested either. (Defensible *iff* a real dockyard/registry key-signing consumer exists — please confirm; otherwise cut to the three symbols.)

## 🟡 Corrected finding (downgraded)

`Bundle.Raw` at `bundles.go:97` uses `json.Marshal(b.bundle)`. A first pass flagged this as not canonical protojson — **that was wrong**: sigstore-go's `bundle.Bundle.MarshalJSON()` *is* `protojson.Marshal` (`pkg/bundle/bundle.go:239`), and `json.Marshal` dispatches to it, so the bytes are canonical. Only a clarity nit: call `b.bundle.MarshalJSON()` explicitly so it doesn't rely on the dispatch.

## 🟡 Confirmed footgun

**`VerifyBundle` default-opts hardcode the public-good instance** (`bundles.go:170-176`). `verifierOptions(PublicGood)` requires SCTs; the GitHub instance does not. A caller passing GitHub trust material while omitting `verifierOpts` gets public-good opts fed to the wrong root → confusing sigstore-go internal error (e.g. "SCTs required but bundle is signed with a public key") instead of a clear opts/material mismatch. Either derive the default from `tm`, or require explicit opts on the general `VerifyBundle` and keep the default only on `VerifyBundleOffline` where the instance is known.

## 🟡 Lifecycle gap

**Frozen `trusted_root.json` has no provenance / freshness signal.** If a Fulcio CA or Rekor key in the snapshot rotates out *due to compromise*, all consumers keep trusting it until a release + bump. Since sigstore-go offers no helper for the target file, keeping the embedded file is right — but please (a) record its provenance (source TUF repo + fetch date/commit), (b) add a CI/`task` drift-check against a freshly-fetched root, and (c) document the compromise-rotation case on `OfflineTrustedMaterial`, not just the "can't verify newly-rotated keys" case.

## 🔵 Lower-priority

- **Missing `doc.go` / stability declaration.** `container/verifier` has no `# Stability` section and isn't in the CLAUDE.md package table, despite 11 new exported symbols. Every other Alpha/Beta package declares it — please add a `doc.go` marking it Alpha and add it to the table. (This is also the cover for the leaky-abstraction trade-off below.)
- **Leaky sigstore-go types on the happy path.** `Bundle.Parsed *bundle.Bundle` (`bundles.go:37`) and `*verify.VerificationResult` returned by all `VerifyBundle*` force every consumer to import sigstore-go — which the `Identity`/`Raw` abstractions otherwise aim to avoid. Acceptable under Alpha, but consider an `Identity`-returning convenience (or an `Example*` showing retrieve → store `Raw` → re-verify offline → `IdentityFromResult`) so the common path doesn't require sigstore-go imports.
- **Error-chain drop:** `RetrieveBundles` converts `ErrProvenanceNotFoundOrIncomplete`→`ErrNoBundles` and drops the chain (`bundles.go:85-86`), breaking `errors.Is(…, ErrProvenanceNotFoundOrIncomplete)` for callers. Consider wrapping instead.
- **No sentinel for "verification failed"** — only `ErrNoBundles` is branchable; everything else is sigstore-go's (unstable) error strings. A typed `ErrVerificationFailed` would let consumers distinguish "signed but failed" from "malformed input".
- **go.mod dep growth is undiscussed.** The `sigstore/sigstore` indirect→direct promotion is mechanically required by `PublicKeyMaterial` (cryptoutils/signature), but it drags ~12 new indirect deps (cobra, pflag, go-tuf, pkg/browser, …). CLAUDE.md treats go.mod changes as needing discussion — worth a note in the PR body.
- Unbounded `io.ReadAll` on the attestation referrer layer (`attestations.go:119`) — no size cap (the sig-manifest path caps at `MaxAttestationsBytesLimit`).

---

**Net:** the security core is sound and correctly uses sigstore-go. Before merge I'd want the **annotation panic** fixed and the **#176 signature alignment + scope question** resolved; the rest is hardening/ergonomics. Nice work on the in-policy identity binding and the clean reuse of the existing internals.

samuv:
Thanks for the thorough panel review — all actionable items addressed in 5127935. Point-by-point:

## 🔴 Ship-blocker — fixed
Every type assertion in `getVerificationMaterialTlogEntries` is now two-valued (the `Payload` map is extracted once with an `ok` check; `apiVersion`/`kind` likewise), returning the existing error path on malformed shapes. Added `sigstore_test.go` with a malformed-annotation matrix (`require.NotPanics` on 10 shapes that previously hit single-value assertions, plus a well-formed control case).

## 🟠 Spec mismatch — aligned where it helps, justified where it deviates
1. **ctx**: `RetrieveBundles(ctx, imageRef, keychain)` now threads context through the entire retrieval chain (`remote.WithContext`/`crane.WithContext` on every registry fetch). The pre-existing exported `GetVerificationResults`/`VerifyServer` keep their signatures (backward compat) and pass `context.Background()` internally, with a comment pointing at `RetrieveBundles` as the cancellable entry point.
2. **`VerifyBundleOffline`** now takes the combined `artifactDigest string` (`"sha256:<hex>"`) per spec. It stays package-level rather than a `*Sigstore` method deliberately: `Sigstore` is the *online* TUF-refreshed verifier, and hanging an offline function off it would imply the instance's live root is involved when it isn't. The `expected *Identity` param is the one intentional superset over #176 — it's what makes identity pinning enforced *inside* the sigstore policy (`WithCertificateIdentity`) rather than post-hoc, which your review verified as the right mechanism.
3. **`IdentityFromResult`** returns an `Identity` value now.
4. **Scope**: real consumers exist for the key flow — stacklok/toolhive#6023 (merged) implements cosign key-pair signing for skills, and the RFC THV-0080 install-verification work consumes `VerifyBundleWithKey` for the signed round-trip and `VerifyBundle`+`expected` for TOFU enforcement of lock-recorded identities (stacklok/toolhive#5899). Happy to link the RFC section if useful.

## 🟡 Footgun — resolved by requiring explicit opts
`VerifyBundle` now errors when no verifier options are passed, with a message pointing at the two valid pairings; `DefaultVerifierOptions()` exposes the public-good set for callers using `OfflineTrustedMaterial`/the live root, and `VerifyBundleOffline` keeps applying it internally since the instance is known there. Guarded by a test.

## 🟡 Lifecycle — documented; drift-check proposed as follow-up
Added `tufroots/README.md` recording provenance (fetched 2026-07-24 from `tuf-repo-cdn.sigstore.dev` via sigstore-go's TUF client, which verifies the target against the TUF metadata chain) and the compromise-rotation staleness in both directions; `OfflineTrustedMaterial`'s doc comment now spells out the compromise case explicitly. A CI drift-check needs network in CI and a refresh policy decision, so I'd prefer to do it as a follow-up issue rather than grow this PR — will file it on merge unless you want it in here.

## 🔵 Lower-priority — all taken except one
- `doc.go` added (Alpha, with a retrieve → verify → store → re-verify-offline example in the package doc) + CLAUDE.md table row.
- `Bundle.Raw` now calls `b.bundle.MarshalJSON()` explicitly.
- `ErrNoBundles` mapping now wraps (`%w: %w`), preserving `errors.Is(…, ErrProvenanceNotFoundOrIncomplete)`.
- `ErrVerificationFailed` sentinel added, wrapping all `Verify` failures in `VerifyBundle`/`VerifyBundleWithKey` (and hence `VerifyBundleOffline`); tests assert `errors.Is` both positively and negatively (malformed input is *not* a verification failure).
- Attestation referrer layer read is now capped at `MaxAttestationsBytesLimit`, matching the sig-manifest path.
- go.mod: noted — the `sigstore/sigstore` direct promotion is mechanically forced by `PublicKeyMaterial` (cryptoutils/signature imports); all of it was already in the module graph via sigstore-go, so no new module versions enter the tree, only direct/indirect reclassification.
- Leaky `*bundle.Bundle`/`*verify.VerificationResult` on the happy path: acknowledged as an Alpha trade-off; the package doc now shows the common path needing only `Raw` + `Identity`. If a second consumer appears I'd wrap the result type before Beta.

JAORMX (approved):

PR #196, opened 2026-07-28
feat(audit): support RFC 8693 act delegation chains

JAORMX:
## Summary

Adds typed, additive RFC 8693 `act` delegation-chain support to the `audit` package, so audit events can capture the actor identity and full delegation chain of a delegated token — not just the terminal `sub`. This closes the shared-library side of stacklok/toolhive#6035 ("Alice did X" vs "Alice's agent did X on her behalf").

**This package is intended to be canonical**: toolhive's `pkg/audit`/`pkg/auth` delegation schema (stacklok/toolhive#6046, #6096) will be reconciled to match this one at migration time.

## What changed

- **`DelegatedActor`** — one hop, identified by typed `iss`+`sub` only (OIDC Connect Core §5.7: the (iss, sub) pair is the only stable, unique actor identifier; RFC 8693 §6 data minimization). All other per-hop claims are retained in memory only and are **never serialized or logged** (PII guard; NIST AU-3(1)/(3) audit-record minimization); accessible via `Extra()` (shallow copy, documented as must-not-be-logged).
- **`DelegationChain`** — flattened, **outermost-first** (`Chain[0]` = current actor, per RFC 8693 §4.1 the only hop that may drive access control; prior hops informational). Current-actor-first keeps `Chain[0]` stable under truncation. `Truncated`/`Omitted`/`Malformed` are always emitted — an incomplete or non-conformant chain is never silently indistinguishable from a complete, well-formed one.
- **Record-and-flag malformed handling** — `ParseDelegationChain(raw, maxDepth) *DelegationChain` **never fails**. Non-conformant input (non-object `act`, non-object nested `act`, non-string `iss`/`sub`) sets `Malformed` plus a low-cardinality `MalformedReason` enum (`act_not_object`, `nested_act_not_object`, `iss_not_string`, `sub_not_string`) and preserves every hop parsed before the violation. An audit record must be producible for exactly the tokens that violate expectations: dropping the record on malformed attacker-influenceable input is an indicator-blocking primitive (MITRE ATT&CK T1562.006; CWE-223/CWE-778), and per RFC 8693 §4.1 prior actors are informational-only, so a malformed chain cannot legitimately fail authorization — recording it is its only purpose. Strict rejection belongs on the token-issuance path (see stacklok/toolhive#6093).
- **`Chain` is always allocated** — the wire always carries `"chain": []`, never `"chain": null` (guarded by a test asserting emitted bytes).
- **Named map types** (e.g. `jwt.MapClaims`) are accepted at any nesting level via a reflection fallback, so programmatically built claims don't silently parse as non-objects.
- **`AuditEvent.DelegationChain`** (`json:"delegation,omitempty"`) + **`WithDelegationChain`** builder. Zero chains (no hops, no truncation, no malformation) are cleared; hopless-but-malformed chains are kept. `Subjects` is **untouched** — the delegating end user is the token's top-level `sub` in `Subjects`, not a chain entry.
- **`LogTo` emits the chain via its JSON marshaling** (`slog.Any`), so the slog and JSON wire shapes of the same event are structurally identical — one parser serves both (guarded by a shape-equality test).
- Fixed the pre-existing `doc.go` bug and added a *Delegation Chains* section documenting ordering, the PII stance, and the malformed-input policy.

## Backward compatibility

**Purely additive — no breaking changes.** One additive struct field with `omitempty`; no constructor/signature changes; `Subjects` contract unchanged. With no chain set, JSON and log output are **byte-identical to before** (guarded by tests). No new dependencies (stdlib only).

## Design decisions

- **Flag, not error.** Evidence: NIST SP 800-53 AU-3(f) (records must identify subjects, no "only if parseable" qualifier) + AU-5 (alert, don't halt); OWASP ASVS 5.0 16.5.3/16.5.4 (fail closed on the *operation*, never lose the error details from the *log*); RFC 9413 (forbids *silent* tolerance — a loud flag satisfies it, discarding violates it); syslog RFC 3164 §4.3.3 record-and-flag precedent; CloudTrail's first-class `userIdentity.type: Unknown`. The reason field is an enum, not free text, per OTel `error.type` low-cardinality guidance, so it is safe as a metric/alert dimension.
- **Depth default 16 (not 10):** this is a *parse-time defensive ceiling* for a shared library, deliberately above any consumer's minting cap, so that for a consumer passing its mint cap (toolhive: 10; draft-mw-spice-actor-chain §21.9 RECOMMENDED default: 10), `truncated: true` is a genuine signal that a token came from outside its own issuance path.
- Chain type lives **directly in the `audit` package**, auth-agnostic — `ParseDelegationChain` takes an already-decoded `any`, no JWT dependency.
- No auto-mirroring of the actor into `Subjects` (avoids smuggled side effects in builders).

## Test plan

`task` green (golangci-lint 0 issues, `go vet` clean, race tests pass); `task license-check` passes. Audit package coverage **99.0%**. Tests assert **emitted JSON keys, not Go struct fields** (tag renames cannot pass with a green suite); pin `"chain":[]` vs `null`; verify slog/JSON shape equality; cover nested parse (0/1/N hops), outermost-first ordering, exactly-at-cap / cap+1 / depth-1000 truncation with exact omitted counts, malformed-past-cap semantics, non-string `sub`/`iss` (flag + in-memory retention), named map types, first-reason-wins, extra-claims PII guards in both JSON and slog output, JSON round-trip, and omit-empty/backward-compat guards.

Refs stacklok/toolhive#6035

### DIFF
--- a/audit/delegation.go
+++ b/audit/delegation.go
@@ -68,6 +68,15 @@ type DelegatedActor struct {
 	extra   map[string]any
 }
 
+// String implements [fmt.Stringer] so that fmt-based rendering of a hop — or of
+// any struct containing hops, such as a dereferenced DelegationChain — cannot
+// reach the unexported extra claims through %v/%+v. fmt does not consult
+// [log/slog.LogValuer], so the PII guard needs this second, fmt-side door
+// closed as well.
+func (a DelegatedActor) String() string {
+	return "{iss:" + a.Issuer + " sub:" + a.Subject + "}"
+}
+
 // Extra returns a shallow copy of the hop's extra (non-standard) claims.
 // Mutating the returned map (adding, deleting, or replacing top-level keys)
 // does not affect the hop's internal state. However, mutating a nested map
@@ -139,6 +148,9 @@ func (c *DelegationChain) IsZero() bool {
 // struct's JSON tags (including omitempty on iss/sub and malformedReason) so
 // slog and JSON output stay structurally identical.
 func (c *DelegationChain) LogValue() slog.Value {
+	if c == nil {
+		return slog.GroupValue()
+	}
 	hops := make([]map[string]string, 0, len(c.Chain))
 	for _, hop := range c.Chain {
 		// Mirror the omitempty tags on DelegatedActor so the shapes stay
--- a/audit/delegation.go
+++ b/audit/delegation.go
@@ -4,26 +4,18 @@
 package audit
 
 import (
-	"encoding/json"
-	"errors"
-	"fmt"
+	"reflect"
 )
 
 // DefaultMaxDelegationDepth is the default defensive cap on delegation-chain
-// parsing. It bounds recursion on attacker-influenceable input: RFC 8693 sets
-// no limit on chain length, so a maliciously crafted token could otherwise
-// force unbounded recursion. Callers may pass a stricter maxDepth to
-// [ParseDelegationChain] to match a minting-side cap. The default is
-// intentionally larger than any known consumer's minting cap.
+// parsing. It bounds the work done on attacker-influenceable input: RFC 8693
+// sets no limit on chain length, so a maliciously crafted token could
+// otherwise force unbounded work. Callers should pass their own minting-side
+// cap as maxDepth to [ParseDelegationChain] when they have one; the default is
+// intentionally larger than any known consumer's minting cap, so that for
+// consumers passing their (smaller) mint cap, Truncated=true is a genuine
+// signal that a token came from outside their own issuance path.
 const DefaultMaxDelegationDepth = 16
-
-// ErrNotJSONObject is returned by [ParseDelegationChain] when the top-level
-// act claim value is not a JSON object (e.g. a string, number, bool, or array).
-var ErrNotJSONObject = errors.New("top-level act claim is not a JSON object")
-
-// ErrNestedActNotObject is returned by [ParseDelegationChain] when a nested
-// act member inside a delegation hop is not a JSON object.
-var ErrNestedActNotObject = errors.New("nested act claim is not a JSON object")
 
 // hopClaimIss, hopClaimSub, and hopClaimAct are the RFC 8693 claim keys
 // recognized on a delegation hop. All other keys are treated as extra
@@ -34,12 +26,39 @@ const (
 	hopClaimAct = "act"
 )
 
+// MalformedReason is a low-cardinality label describing the first
+// RFC 8693 conformance violation encountered while parsing an `act` claim.
+// It is deliberately an enum rather than free text so that it is safe to use
+// as a metric or alerting dimension.
+type MalformedReason string
+
+// Malformed reasons reported by [ParseDelegationChain]. The first violation
+// encountered wins; subsequent violations do not overwrite it.
+const (
+	// MalformedReasonActNotObject: the top-level act claim value was not a
+	// JSON object (e.g. a string, number, bool, or array).
+	MalformedReasonActNotObject MalformedReason = "act_not_object"
+	// MalformedReasonNestedActNotObject: a nested act member inside a
+	// delegation hop was not a JSON object.
+	MalformedReasonNestedActNotObject MalformedReason = "nested_act_not_object"
+	// MalformedReasonIssNotString: a hop carried an iss claim that was not a
+	// JSON string. The raw value is retained in that hop's in-memory extra
+	// claims under "iss".
+	MalformedReasonIssNotString MalformedReason = "iss_not_string"
+	// MalformedReasonSubNotString: a hop carried a sub claim that was not a
+	// JSON string. The raw value is retained in that hop's in-memory extra
+	// claims under "sub".
+	MalformedReasonSubNotString MalformedReason = "sub_not_string"
+)
+
 // DelegatedActor represents a single hop in an RFC 8693 delegation chain.
 //
-// Per-hop identity is the issuer (iss) and subject (sub) pair only, as
-// specified by RFC 8693 §6. Any other claims present on a hop are retained
-// in the in-memory-only extra map (never serialized or logged) to avoid
-// leaking Personally Identifiable Information.
+// Per-hop identity is the issuer (iss) and subject (sub) pair only: per
+// OpenID Connect Core §5.7 the (iss, sub) combination is the only stable,
+// unique actor identifier, and RFC 8693 §6 calls for data minimization. Any
+// other claims present on a hop are retained in the in-memory-only extra map
+// (never serialized or logged) to avoid leaking Personally Identifiable
+// Information.
 type DelegatedActor struct {
 	Issuer  string `json:"iss,omitempty"`
 	Subject string `json:"sub,omitempty"`
@@ -51,6 +70,11 @@ type DelegatedActor struct {
 // does not affect the hop's internal state. However, mutating a nested map
 // or slice value may share underlying data with the hop's copy. It returns
 // nil when the hop has no extra claims.
+//
+// The extra claims are deliberately excluded from JSON and slog output as a
+// PII guard, and may contain sensitive or internal claims (session
+// identifiers, emails, roles). Callers MUST NOT serialize or log the returned
+// map; doing so reintroduces the leak this design prevents.
 func (a DelegatedActor) Extra() map[string]any {
 	if len(a.extra) == 0 {
 		return nil
@@ -65,158 +89,206 @@ func (a DelegatedActor) Extra() map[string]any {
 // DelegationChain holds a flattened, ordered RFC 8693 delegation chain.
 //
 // The chain is ordered OUTERMOST-FIRST: Chain[0] is the current actor (the
-// immediate subject of the event), and the last entry is the least recent
-// (original) actor. Only the top-level event claims and Chain[0] should
-// drive access-control decisions; prior hops are informational only.
+// immediate actor of the event), and the last entry is the least recent
+// (earliest) actor. The delegating end user is NOT part of the chain: per
+// RFC 8693 it is the token's top-level sub, recorded in the event's Subjects
+// map. Per RFC 8693 §4.1, only the top-level event claims and Chain[0] may
+// drive access-control decisions; prior hops are informational only and exist
+// for audit.
 //
-// Truncated and Omitted intentionally have no omitempty tag: truncation is
-// never silent. When Truncated is true, Omitted reports how many inner hops
-// were dropped to satisfy the maxDepth cap. The outermost hops are always
-// preserved (the current actor is never dropped); only inner hops are
-// omitted.
+// Ordering rationale — do not assume, and do not "fix": current-actor-first
+// matches the read order of RFC 8693's nested act structure, and it keeps
+// Chain[0] STABLE UNDER TRUNCATION — truncation drops inner (early) hops, so
+// the one hop that may inform decisions is always at index 0, complete chain
+// or not. Beware that append-style flat formats elsewhere (Envoy XFCC,
+// GCP delegates[]) run the OPPOSITE direction (original-first); consumers
+// correlating across systems must map explicitly rather than by position
+// intuition.
+//
+// Truncated, Omitted, and Malformed intentionally have no omitempty tag:
+// an incomplete or non-conformant chain is never silently indistinguishable
+// from a complete, well-formed one. When Truncated is true, Omitted reports
+// how many inner hops were dropped to satisfy the maxDepth cap; the outermost
+// hops are always preserved (the current actor is never dropped). When
+// Malformed is true, MalformedReason carries the first conformance violation
+// encountered, and Chain holds the hops that parsed successfully before it.
 type DelegationChain struct {
-	Chain     []DelegatedActor `json:"chain"`
-	Truncated bool             `json:"truncated"`
-	Omitted   int              `json:"omitted"`
+	Chain           []DelegatedActor `json:"chain"`
+	Truncated       bool             `json:"truncated"`
+	Omitted         int              `json:"omitted"`
+	Malformed       bool             `json:"malformed"`
+	MalformedReason MalformedReason  `json:"malformedReason,omitempty"`
 }
 
-// IsEmpty reports whether the chain is nil or contains no hops.
+// IsEmpty reports whether the chain is nil or contains no hops. Note that a
+// chain with no hops may still carry audit-relevant state (Malformed); use
+// [DelegationChain.IsZero] to decide whether a chain is worth recording.
 func (c *DelegationChain) IsEmpty() bool {
 	return c == nil || len(c.Chain) == 0
+}
+
+// IsZero reports whether the chain carries no information at all: no hops,
+// no truncation, and no malformation. A zero chain is omitted from JSON and
+// log output; a non-zero chain — including a hopless but malformed one — is
+// always recorded.
+func (c *DelegationChain) IsZero() bool {
+	return c == nil || (len(c.Chain) == 0 && !c.Truncated && !c.Malformed)
+}
+
+// flagMalformed marks the chain malformed with the given reason. The first
+// reason encountered wins so that the recorded reason points at the outermost
+// (most actionable) violation.
+func (c *DelegationChain) flagMalformed(reason MalformedReason) {
+	c.Malformed = true
+	if c.MalformedReason == "" {
+		c.MalformedReason = reason
+	}
 }
 
 // ParseDelegationChain parses a raw RFC 8693 `act` claim into a flattened,
 // outermost-first [DelegationChain].
 //
-// The input is expected to be a JSON object (map) decoded via
-// [encoding/json.Unmarshal] into an `any` value (i.e. map[string]any). The
-// parser walks nested `act` claims iteratively, bounded by maxDepth, so
-// pathologically deep input cannot cause a stack overflow.
+// The input is expected to be the decoded JSON value of the `act` claim
+// (i.e. a map[string]any, or any named map type with string keys such as
+// jwt.MapClaims). The parser walks nested `act` claims iteratively, bounded
+// by maxDepth, so pathologically deep input cannot cause a stack overflow.
+//
+// ParseDelegationChain never fails: it is an audit-side parser, and an audit
+// record must be producible for exactly the tokens that violate expectations
+// (an unreadable delegation assertion is forensically distinct from "no
+// delegation"). Non-conformant input is recorded on the returned chain via
+// Malformed/MalformedReason rather than dropped or returned as an error.
+// Enforcement belongs on the issuance path, not here.
 //
 // Parsing semantics:
 //   - maxDepth <= 0 uses [DefaultMaxDelegationDepth].
-//   - A nil input yields a non-nil empty chain (&DelegationChain{}), no error.
-//   - A non-map scalar (string, float64, bool, []any) returns nil and an error
-//     whose message names the unexpected kind.
+//   - A nil input yields a zero chain (no hops, not malformed): a JSON null
+//     or absent act claim asserts no delegation.
+//   - A non-object input (string, number, bool, array) yields no hops and
+//     Malformed=true with [MalformedReasonActNotObject].
 //   - A map with no iss/sub/act yields one hop with empty Subject/Issuer
 //     (sub is conventional, not RFC-mandatory; parsing is tolerant).
-//   - A map with sub sets that hop's Subject; iss sets Issuer.
+//   - A string sub sets that hop's Subject; a string iss sets Issuer. A
+//     non-string sub or iss leaves the field empty, flags the chain malformed,
+//     and retains the raw value in the hop's in-memory extra claims — the
+//     record shows that an identifier was present but unusable.
 //   - A nested `act` map is recursed: the outer map is Chain[0], its act is
 //     Chain[1], and so on. The result is naturally outermost-first.
+//   - A nested `act` that is present but not a map (JSON null excepted, which
+//     ends the chain) keeps the hops parsed so far and flags the chain
+//     malformed with [MalformedReasonNestedActNotObject].
 //   - When the depth equals maxDepth exactly, the full chain is returned with
 //     Truncated=false and Omitted=0.
 //   - When the depth exceeds maxDepth, the OUTERMOST maxDepth hops are kept
 //     (the current actor is preserved), Truncated=true, and Omitted counts
-//     the dropped inner hops. This is truncation, not an error.
-//   - An `act` claim present but not a map (e.g. "act":"x") returns nil and
-//     an error.
+//     the dropped inner hops. A malformed node past the cap flags Malformed
+//     but is not counted as an omitted hop.
 //   - Unknown extra keys on a hop are stored in that hop's in-memory-only
-//     extra map (see [DelegatedActor.Extra]).
-func ParseDelegationChain(raw any, maxDepth int) (*DelegationChain, error) {
+//     extra map (see [DelegatedActor.Extra]) and are never serialized.
+func ParseDelegationChain(raw any, maxDepth int) *DelegationChain {
 	if maxDepth <= 0 {
 		maxDepth = DefaultMaxDelegationDepth
 	}
 
+	chain := &DelegationChain{Chain: []DelegatedActor{}}
 	if raw == nil {
-		return &DelegationChain{}, nil
+		return chain
 	}
 
-	// A non-map scalar input is invalid; the top-level act claim must be a
-	// JSON object. Reject by naming the unexpected kind.
-	switch raw.(type) {
-	case map[string]any:
-		// ok, proceed
-	case string, float64, bool, []any, json.Number:
-		return nil, fmt.Errorf("%w: expected JSON object, got %T", ErrNotJSONObject, raw)
-	default:
-		return nil, fmt.Errorf("%w: expected JSON object, got %T", ErrNotJSONObject, raw)
+	current, ok := asStringMap(raw)
+	if !ok {
+		chain.flagMalformed(MalformedReasonActNotObject)
+		return chain
 	}
 
-	chain := &DelegationChain{}
-	current := raw
-	depth := 0
-
-	for current != nil {
+	for depth := 0; ; depth++ {
 		if depth >= maxDepth {
 			// We have hit the cap. The remaining inner hops (current and any
-			// nested act beneath it) must be counted and dropped, preserving
-			// only the outermost maxDepth hops already collected.
+			// nested act beneath it) are counted and dropped, preserving only
+			// the outermost maxDepth hops already collected.
 			chain.Truncated = true
-			chain.Omitted += countRemainingHops(current)
-			return chain, nil
+			omitted, malformed := countRemainingHops(current)
+			chain.Omitted += omitted
+			if malformed {
+				chain.flagMalformed(MalformedReasonNestedActNotObject)
+			}
+			return chain
 		}
 
-		hopMap, ok := current.(map[string]any)
-		if !ok {
-			// An `act` claim present but not a map is a hard error.
-			return nil, fmt.Errorf("%w: act claim must be a JSON object", ErrNestedActNotObject)
+		hop := DelegatedActor{extra: extraClaims(current)}
+		if rawIss, present := current[hopClaimIss]; present {
+			if iss, isString := rawIss.(string); isString {
+				hop.Issuer = iss
+			} else {
+				chain.flagMalformed(MalformedReasonIssNotString)
+			}
 		}
-
-		hop := DelegatedActor{extra: extraClaims(hopMap)}
-		if iss, ok := hopMap[hopClaimIss].(string); ok {
-			hop.Issuer = iss
-		}
-		if sub, ok := hopMap[hopClaimSub].(string); ok {
-			hop.Subject = sub
+		if rawSub, present := current[hopClaimSub]; present {
+			if sub, isString := rawSub.(string); isString {
+				hop.Subject = sub
+			} else {
+				chain.flagMalformed(MalformedReasonSubNotString)
+			}
 		}
 		chain.Chain = append(chain.Chain, hop)
-		depth++
 
-		next, hasAct := hopMap[hopClaimAct]
-		if !hasAct {
-			// No further nesting; the chain is complete.
-			return chain, nil
+		next, hasAct := current[hopClaimAct]
+		if !hasAct || next == nil {
+			// No further nesting (an explicit JSON null ends the chain like
+			// an absent act); the chain is complete.
+			return chain
 		}
 
-		nextMap, ok := next.(map[string]any)
+		nextMap, ok := asStringMap(next)
 		if !ok {
-			return nil, fmt.Errorf("%w: act claim must be a JSON object", ErrNestedActNotObject)
+			// A nested act present but not an object: keep what parsed and
+			// record the violation.
+			chain.flagMalformed(MalformedReasonNestedActNotObject)
+			return chain
 		}
 		current = nextMap
 	}
-
-	return chain, nil
 }
 
-// countRemainingHops counts how many hops would follow the current node
-// (inclusive of current) down the nested `act` chain. These are the hops
-// dropped due to exceeding maxDepth. The walk is iterative (not recursive),
-// so it cannot stack-overflow; it terminates because decoded JSON is a tree
-// and cannot contain reference cycles.
-func countRemainingHops(node any) int {
+// countRemainingHops counts how many hops follow the current node (inclusive
+// of current) down the nested `act` chain. These are the hops dropped due to
+// exceeding maxDepth. The second return reports whether the walk ended on a
+// non-object act value. The walk is iterative (not recursive), so it cannot
+// stack-overflow; it terminates because decoded JSON is a tree and cannot
+// contain reference cycles.
+func countRemainingHops(node map[string]any) (int, bool) {
 	count := 0
 	current := node
-	for current != nil {
-		m, ok := current.(map[string]any)
-		if !ok {
-			// Non-map act was already validated during the main walk; treat a
-			// malformed node as a single dropped hop and stop.
-			count++
-			break
-		}
+	for {
 		count++
-		next, hasAct := m[hopClaimAct]
-		if !hasAct {
-			break
+		next, hasAct := current[hopClaimAct]
+		if !hasAct || next == nil {
+			return count, false
 		}
-		nm, ok := next.(map[string]any)
+		nm, ok := asStringMap(next)
 		if !ok {
-			count++
-			break
+			// A non-object act past the cap is a violation, not a hop.
+			return count, true
 		}
 		current = nm
 	}
-	return count
 }
 
-// extraClaims returns a shallow copy of the non-standard keys on a hop map
-// (i.e. every key except iss, sub, and act). The result is nil when there are
-// no extra claims.
+// extraClaims returns a shallow copy of the non-standard keys on a hop map:
+// every key except act, and except iss/sub when they hold their conventional
+// string type. A non-string iss/sub is retained here so the raw value stays
+// inspectable in memory after the typed field is left empty. The result is
+// nil when there are no extra claims.
 func extraClaims(hop map[string]any) map[string]any {
 	var extra map[string]any
 	for k, v := range hop {
-		if k == hopClaimIss || k == hopClaimSub || k == hopClaimAct {
+		switch k {
+		case hopClaimAct:
 			continue
+		case hopClaimIss, hopClaimSub:
+			if _, isString := v.(string); isString {
+				continue
+			}
 		}
 		if extra == nil {
 			extra = make(map[string]any)
@@ -224,4 +296,24 @@ func extraClaims(hop map[string]any) map[string]any {
 		extra[k] = v
 	}
 	return extra
+}
+
+// asStringMap converts a decoded JSON object into a map[string]any. The fast
+// path is a direct type assertion; the reflective fallback accepts named map
+// types with string keys (e.g. jwt.MapClaims) so that programmatically built
+// claim structures are not silently misread as non-objects.
+func asStringMap(v any) (map[string]any, bool) {
+	if m, ok := v.(map[string]any); ok {
+		return m, true
+	}
+	rv := reflect.ValueOf(v)
+	if !rv.IsValid() || rv.Kind() != reflect.Map || rv.Type().Key().Kind() != reflect.String {
+		return nil, false
+	}
+	out := make(map[string]any, rv.Len())
+	iter := rv.MapRange()
+	for iter.Next() {
+		out[iter.Key().String()] = iter.Value().Interface()
+	}
+	return out, true
 }
--- a/audit/delegation.go
+++ b/audit/delegation.go
@@ -7,14 +7,16 @@ import (
 	"reflect"
 )
 
-// DefaultMaxDelegationDepth is the default defensive cap on delegation-chain
-// parsing. It bounds the work done on attacker-influenceable input: RFC 8693
-// sets no limit on chain length, so a maliciously crafted token could
-// otherwise force unbounded work. Callers should pass their own minting-side
-// cap as maxDepth to [ParseDelegationChain] when they have one; the default is
-// intentionally larger than any known consumer's minting cap, so that for
-// consumers passing their (smaller) mint cap, Truncated=true is a genuine
-// signal that a token came from outside their own issuance path.
+// DefaultMaxDelegationDepth is the default cap on how many delegation hops
+// [ParseDelegationChain] retains. It is a parse ceiling, not a minting cap:
+// RFC 8693 sets no limit on chain length, so this bounds the size of the
+// retained chain and of the emitted audit record on attacker-influenceable
+// input. It does not bound parsing work — the caller has already decoded the
+// claim. Callers should pass their own minting-side cap as maxDepth when they
+// have one; the default is intentionally larger than any known consumer's
+// minting cap, so that for consumers passing their (smaller) mint cap,
+// Truncated=true is a genuine signal that a token came from outside their own
+// issuance path.
 const DefaultMaxDelegationDepth = 16
 
 // hopClaimIss, hopClaimSub, and hopClaimAct are the RFC 8693 claim keys
@@ -88,13 +90,13 @@ func (a DelegatedActor) Extra() map[string]any {
 
 // DelegationChain holds a flattened, ordered RFC 8693 delegation chain.
 //
-// The chain is ordered OUTERMOST-FIRST: Chain[0] is the current actor (the
-// immediate actor of the event), and the last entry is the least recent
-// (earliest) actor. The delegating end user is NOT part of the chain: per
-// RFC 8693 it is the token's top-level sub, recorded in the event's Subjects
-// map. Per RFC 8693 §4.1, only the top-level event claims and Chain[0] may
-// drive access-control decisions; prior hops are informational only and exist
-// for audit.
+// The chain is ordered OUTERMOST-FIRST: Chain[0] is the current actor — the
+// party that presented the token — and the last entry is the least recent
+// (original) actor. The party on whose behalf they acted is NOT part of the
+// chain: per RFC 8693 it is the token's top-level sub, carried by the event's
+// own Subjects map; the chain never repeats it. Per RFC 8693 §4.1, only the
+// top-level event claims and Chain[0] may drive access-control decisions;
+// prior hops are informational only and exist for audit.
 //
 // Ordering rationale — do not assume, and do not "fix": current-actor-first
 // matches the read order of RFC 8693's nested act structure, and it keeps
--- a/audit/delegation.go
+++ b/audit/delegation.go
@@ -4,6 +4,7 @@
 package audit
 
 import (
+	"log/slog"
 	"reflect"
 )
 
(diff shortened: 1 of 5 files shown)


jhrozek:
Heads up — this overlaps with work that already merged on the toolhive side, and the two schemas currently disagree on every field name.

stacklok/toolhive#6046 (merged earlier today) implemented the same feature for `pkg/audit`, and stacklok/toolhive#6096 is a follow-up hardening it. Since `audit/event.go` here was graduated out of toolhive in #81, both PRs are adding a delegation field to what is logically the same struct — so whichever schema survives the eventual `pkg/audit` → `core/audit` migration silently rewrites audit log output for anyone keying on the other one.

Where we diverge:

| | toolhive `pkg/audit` + `pkg/auth` | this PR |
|---|---|---|
| event field | `delegationChain,omitempty` | `delegation,omitempty` |
| hop list | `Actors` → `actors` | `Chain` → `chain` |
| dropped counter | `dropped_count,omitempty` | `omitted` (no omitempty) |
| per-hop `iss` | folded into `act_claims` | typed `Issuer` → `iss` |
| other per-hop claims | serialized as `act_claims`, minus internal `tsid` | never serialized (unexported `extra`, `json:"-"`) |
| malformed `act` | tolerant: `Malformed bool` flag, partial chain kept | hard error: `ErrNotJSONObject` / `ErrNestedActNotObject` |
| default depth | 10 | 16 |
| slog shape | one `slog.Any("delegation_chain", chain)` blob | `slog.Group` with `hops.hop_N.{iss,sub}` |

The good news is that the highest-risk thing agrees: **actor ordering is identical** — both outermost-first, both keep the outermost hops on truncation, both report not-truncated when the chain is exactly at the cap. That was the one divergence that would have made audit records actively misleading, and it is not one.

Two things worth deciding rather than discovering later.

**1. The PII question is a real design disagreement, and I think you are right.** We serialize arbitrary per-actor `act` members as `act_claims` and filter only `tsid`; this PR refuses to serialize them at all, citing RFC 8693 §6 data minimization. That is the stronger position. Worth noting the concrete reason it matters on our side: `Identity.MarshalJSON` redacts `tsid` from top-level claims but marshals the delegation chain **verbatim** to admission webhook receivers, so before the `internalClaims` filter in #6096 an `act.tsid` from a foreign IdP would have leaked there. Your design makes that class of bug impossible instead of merely filtered.

**2. Malformed input: flag or error?** You return `ErrNotJSONObject`; we set `malformed: true` and keep whatever hops parsed. Our reasoning was that an audit record should be able to say "the caller presented a token asserting delegation that we could not read", which is distinct from "no delegation" — and because signature and issuer validation run first, a malformed `act` can only come from a trusted issuer, so it flags an issuer bug worth recording rather than client noise. That argument is specific to the audit sink; for a library parser, returning an error is defensible. Either way the two should not disagree.

Concretely relevant to (2): stacklok/toolhive#6093 — the toolhive auth server can itself currently mint a non-conformant nested `act`, because the token-exchange handler copies a prior `act` without checking it is a JSON object, and its depth counter returns 0 for a non-map so the depth gate waves it through. Malformed chains are reachable from our own issuance path, not just hypothetically from third parties.

## Findings from the #6046 review that may apply here too

Listing these so we do not have to find them twice.

- **`Chain` has no `omitempty`, so a nil slice marshals to `"chain": null`.** We hit exactly this: admitting an actorless chain made `"actors": null` reachable where it never had been, which breaks `jq` array iteration and Python `for` loops on precisely the malformed tokens the feature exists to surface. We now always allocate the slice. Your hard-error path may make it unreachable — worth confirming, and pinning either way, since `assert.Empty` passes for both `null` and `[]`.
- **Assert the emitted JSON keys, not just the Go struct fields.** Ours were struct-field-only, so a tag rename would have broken every downstream consumer with a green suite.
- **Check the exactly-at-cap and cap+1 boundaries explicitly**, including the precise dropped/omitted count for a chain much deeper than the cap. Our accumulating `DroppedCount +=` would have passed as a plain `=`, because no test supplied a non-zero incoming count.
- **A non-string `sub`.** We captured `sub` only when it type-asserted to string but dropped the key from the claims copy unconditionally, so `{"sub": 123}` produced a completely empty actor object — an anonymous actor with no trace that an identifier had been present. Your typed `Subject`/`Issuer` fields may have the same shape of hole.
- **A named map type will not type-assert.** `jwt.MapClaims` is `map[string]interface{}`, but a `.(map[string]any)` assertion fails on it. In practice only the top-level claims object is `jwt.MapClaims` — nested values come from `encoding/json` as plain maps — so this is fine on the real path, but a programmatically-built `jwt.MapClaims{"act": jwt.MapClaims{...}}` silently yields nothing.
- **Cover every emitter, not a representative one.** Our audit type repeats the attach call in eight methods and only two were tested; the six uncovered ones included every terminal failed/timed-out event, so dropping the call from any of them would have shipped silently.
- **If both keep a depth default, reconcile 10 vs 16 or document the distinction** — parse ceiling vs mint cap. The toolhive mint cap is 10, so a locally-issued chain never truncates, which makes `truncated: true` a genuine signal that a token came from elsewhere. That property is worth preserving deliberately rather than by accident.

Happy to change ours if you would rather core be canonical — nothing outside toolhive depends on our field names yet, so now is the cheapest moment. What I would like to avoid is both shipping and then reconciling after someone has written alerts against one of them.

audit/delegation.go
```diff
@@ -0,0 +1,227 @@
...
+	"errors"
+	"fmt"
+)
+
+// DefaultMaxDelegationDepth is the default defensive cap on delegation-chain
+// parsing. It bounds recursion on attacker-influenceable input: RFC 8693 sets
+// no limit on chain length, so a maliciously crafted token could otherwise
+// force unbounded recursion. Callers may pass a stricter maxDepth to
+// [ParseDelegationChain] to match a minting-side cap. The default is
+// intentionally larger than any known consumer's minting cap.
```
jhrozek: The DoS rationale oversells what the cap does. By the time this runs the caller has already decoded the claim, so `encoding/json` did the deep recursion (under its own nesting limit); this walk is iterative; and `countRemainingHops` traverses the entire remaining chain regardless of the cap. The cap bounds the size of the retained chain and the emitted record — not work, not stack. Also worth saying outright that 16 is a *parse ceiling* distinct from a minting cap, since toolhive#6046 chose 10 and the two numbers will get compared.

```suggestion
// DefaultMaxDelegationDepth is the default cap on how many delegation hops
// [ParseDelegationChain] retains. It is a parse ceiling, not a minting cap:
// RFC 8693 sets no limit on chain length, so this bounds the size of the
// retained chain and of the emitted audit record on attacker-influenceable
// input. It does not bound parsing work — the caller has already decoded the
// claim. Callers may pass a stricter maxDepth to match a minting-side cap;
// the default is intentionally larger than any known consumer's.
```

audit/delegation.go
```diff
@@ -0,0 +1,227 @@
...
+// Truncated and Omitted intentionally have no omitempty tag: truncation is
+// never silent. When Truncated is true, Omitted reports how many inner hops
+// were dropped to satisfy the maxDepth cap. The outermost hops are always
+// preserved (the current actor is never dropped); only inner hops are
+// omitted.
+type DelegationChain struct {
+	Chain     []DelegatedActor `json:"chain"`
+	Truncated bool             `json:"truncated"`
+	Omitted   int              `json:"omitted"`
+}
```
jhrozek, line 133: Adds `Malformed` (see the review body), and two doc fixes: *"Chain[0] is the current actor (the immediate subject of the event)"* conflates the two parties RFC 8693 4.1 keeps apart — `sub` is who the action is taken on behalf of, `act` is who is acting — and the chain never carries the top-level `sub` at all, so a consumer reading only `delegation` cannot tell on whose behalf anything happened.

```suggestion
// DelegationChain holds a flattened, ordered RFC 8693 delegation chain.
//
// The chain is ordered OUTERMOST-FIRST: Chain[0] is the current actor — the
// party that presented the token — and the last entry is the least recent
// (original) actor. The party on whose behalf they acted is carried by the
// event's own Subjects map; the chain never repeats it. Per RFC 8693 4.1,
// only the top-level claims and Chain[0] may drive access-control decisions;
// prior hops are informational only.
//
// Truncated, Omitted, and Malformed intentionally have no omitempty tag:
// neither truncation nor a parse failure is ever silent. When Truncated is
// true, Omitted reports how many inner hops were dropped to satisfy the
// maxDepth cap; the outermost hops are always preserved, so the current actor
// is never dropped. Malformed reports that parsing stopped at a
// non-conformant `act`; the hops parsed before that point are still present.
type DelegationChain struct {
	Chain     []DelegatedActor `json:"chain"`
	Truncated bool             `json:"truncated"`
	Omitted   int              `json:"omitted"`
	Malformed bool             `json:"malformed"`
}
```

audit/delegation.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+		nm, ok := next.(map[string]any)
+		if !ok {
+			count++
+			break
+		}
+		current = nm
+	}
+	return count
+}
```
jhrozek, line 316: The main change. Never return a nil chain; flag `Malformed` instead; always allocate `Chain` so it marshals to `[]`; treat JSON `null` as chain-terminating; count only objects in `countRemainingHops`; and carry `map[string]any` through the loop so the four unreachable branches disappear (`:129-133` were an identical `case`/`default`, `:149-153` and `:178` could not be reached, `:191-195` likewise — together they are every uncovered statement in the file).

Applied on this branch: `task lint` clean, `go test -race ./audit/` green, coverage 91.3% → 98.7%, 27 lines shorter.

```suggestion
// ParseDelegationChain parses a raw RFC 8693 `act` claim into a flattened,
// outermost-first [DelegationChain].
//
// The input is expected to be a JSON object (map) decoded via
// [encoding/json.Unmarshal] into an `any` value (i.e. map[string]any). The
// parser walks nested `act` claims iteratively, bounded by maxDepth, so
// pathologically deep input cannot cause a stack overflow.
//
// Parsing semantics:
//   - maxDepth <= 0 uses [DefaultMaxDelegationDepth].
//   - The returned chain is ALWAYS non-nil, including alongside an error, so a
//     caller that records the event still records that delegation was
//     asserted. Chain is always allocated, so it marshals to [], never null.
//   - A nil input yields an empty chain and no error.
//   - A non-map input, or a nested `act` that is present but not a map, sets
//     Malformed, returns the hops parsed so far, and reports an error naming
//     the unexpected kind ([ErrNotJSONObject] / [ErrNestedActNotObject]).
//   - An absent `act`, or one whose value is JSON null, ends the chain. RFC
//     8693 4.1 requires a *present* act to be an object; null is not
//     non-conformant.
//   - A map with no iss/sub yields one hop with empty Issuer/Subject (sub is
//     conventional, not RFC-mandatory; parsing is tolerant).
//   - iss and sub set the hop's Issuer and Subject when they are strings. A
//     non-string iss/sub is invalid per RFC 7519 4.1.2 and is preserved in the
//     hop's extra map rather than discarded, since a malformed identifier is
//     itself a signal about the issuer.
//   - A nested `act` map is flattened: the outer map is Chain[0], its act is
//     Chain[1], and so on. The result is naturally outermost-first.
//   - When the depth equals maxDepth exactly, the full chain is returned with
//     Truncated=false and Omitted=0.
//   - When the depth exceeds maxDepth, the OUTERMOST maxDepth hops are kept
//     (the current actor is preserved), Truncated=true, and Omitted counts the
//     dropped inner hops. Only JSON objects count. This is truncation, not an
//     error.
//   - Extra keys on a hop are stored in that hop's in-memory-only extra map
//     (see [DelegatedActor.Extra]).
func ParseDelegationChain(raw any, maxDepth int) (*DelegationChain, error) {
	if maxDepth <= 0 {
		maxDepth = DefaultMaxDelegationDepth
	}

	// Chain is always allocated: a nil slice would marshal to "chain": null.
	chain := &DelegationChain{Chain: []DelegatedActor{}}
	if raw == nil {
		return chain, nil
	}

	hopMap, ok := raw.(map[string]any)
	if !ok {
		chain.Malformed = true
		return chain, fmt.Errorf("%w: got %T", ErrNotJSONObject, raw)
	}

	for depth := 0; ; depth++ {
		if depth >= maxDepth {
			chain.Truncated = true
			chain.Omitted = countRemainingHops(hopMap)
			return chain, nil
		}

		hop := DelegatedActor{extra: extraClaims(hopMap)}
		if iss, ok := hopMap[hopClaimIss].(string); ok {
			hop.Issuer = iss
		}
		if sub, ok := hopMap[hopClaimSub].(string); ok {
			hop.Subject = sub
		}
		chain.Chain = append(chain.Chain, hop)

		// An absent act ends the chain; so does JSON null, which RFC 8693 4.1
		// does not treat as non-conformant.
		next := hopMap[hopClaimAct]
		if next == nil {
			return chain, nil
		}
		if hopMap, ok = next.(map[string]any); !ok {
			// Non-conformant act from an already-trust-validated issuer. Keep
			// the hops parsed so far and flag; never discard the record.
			chain.Malformed = true
			return chain, fmt.Errorf("%w: got %T", ErrNestedActNotObject, next)
		}
	}
}

// countRemainingHops counts the hops at and below node that were dropped to
// satisfy maxDepth. Only JSON objects count: a non-object act is not an actor.
// The walk is iterative, so it cannot stack-overflow, and it terminates
// because decoded JSON is a tree and cannot contain reference cycles.
func countRemainingHops(node any) int {
	count := 0
	for {
		m, ok := node.(map[string]any)
		if !ok {
			return count
		}
		count++
		node = m[hopClaimAct]
	}
}
```

audit/delegation.go
```diff
@@ -0,0 +1,227 @@
...
+
+// extraClaims returns a shallow copy of the non-standard keys on a hop map
+// (i.e. every key except iss, sub, and act). The result is nil when there are
+// no extra claims.
+func extraClaims(hop map[string]any) map[string]any {
+	var extra map[string]any
+	for k, v := range hop {
+		if k == hopClaimIss || k == hopClaimSub || k == hopClaimAct {
+			continue
+		}
```
jhrozek: `iss`/`sub` are skipped unconditionally, but the typed fields only capture them when they are strings — so `{"sub": 123}` produces a hop that is byte-identical to `{}` in the audit record, with `Extra()` nil and the raw value nowhere. RFC 7519 4.1.2 makes a numeric `sub` invalid, but a malformed identifier from a trust-validated issuer is evidence, not noise. This keeps it inspectable without putting it on the wire.

```suggestion
// extraClaims returns a shallow copy of the claims on a hop map that no typed
// field captures: every key except act, plus iss/sub when they are present but
// not strings (invalid per RFC 7519 4.1.2, retained as evidence about the
// issuer). The result is nil when there are no such claims.
func extraClaims(hop map[string]any) map[string]any {
	var extra map[string]any
	for k, v := range hop {
		if k == hopClaimAct {
			continue
		}
		if k == hopClaimIss || k == hopClaimSub {
			if _, isString := v.(string); isString {
				continue // captured by the typed field
			}
		}
```

jhrozek (commented):
## Overlaps a merged toolhive PR, and the hard-error contract is doing damage

Two things, then four one-click suggestions that I applied and verified on this branch before proposing them.

**1. Schema decision (cross-repo, not code).** toolhive#6046 merged this same feature earlier today with a different wire schema:

| | toolhive (merged) | this PR |
|---|---|---|
| event field | `delegationChain` | `delegation` |
| hop list | `actors` | `chain` |
| dropped counter | `dropped_count,omitempty` | `omitted` |
| per-hop `iss` | folded into `act_claims` | typed `iss` |
| other per-hop claims | serialized as `act_claims` | never serialized |
| malformed `act` | `Malformed bool`, partial chain | hard error |
| default depth | 10 | 16 |
| slog shape | one `slog.Any` blob | `slog.Group`, `hops.hop_N` |

No log consumer can be written against both. Ordering and truncation direction already agree, so this is names plus one design call. Core's `audit` is toolhive's `pkg/audit` graduated (core#81), so core's keys win by construction and toolhive's emitted keys will change silently under existing consumers when that migration lands. Nothing outside toolhive depends on either set yet — cheapest moment to converge is now.

On the one substantive difference, **your unexported `extra` should win.** toolhive marshals the chain verbatim to admission-webhook receivers, so a foreign `act.tsid` reached third parties unredacted until #6096 added a filter. Yours makes that class of bug unrepresentable rather than filtered.

**2. `(nil, err)` is the root of most of what I found.** Per toolhive#6093, our own token-exchange handler copies a prior `act` without checking it is an object (`actChainDepth` returns 0 for a non-map, so the depth gate waves it through) — malformed chains are reachable from our own AS. Because signature and issuer validation run first, a malformed `act` can only arrive from a **trusted** issuer, so it flags an issuer bug, which is exactly what an audit trail exists to capture. Returning a nil chain means it vanishes from the record, and the record cannot distinguish "no delegated token" from "token asserted delegation we could not read".

Adopting toolhive's `Malformed bool` and never returning a nil chain closes six findings at once:

| finding | before | after |
|---|---|---|
| chain vanishes on malformed `act` from a trusted issuer | `(nil, err)` | partial chain + `Malformed: true` |
| **depth invariance broken** — `h1→h2→h3→act:"garbage"` yields 2 hops / `omitted=2` at `maxDepth=2`, but a hard error at `maxDepth=16` | record vs nothing, operator's knob decides | same record either way |
| `omitted` counted a bare string as a dropped hop | `omitted=2` for one real dropped hop | objects only |
| `json.Marshal(ParseDelegationChain(nil, 0))` → `{"chain":null,…}`, breaking `jq` iteration | `null` | `[]` |
| `{"sub":"h1","act":null}` errored, though 4.1 only makes *present-but-not-an-object* non-conformant | error | terminates the chain |
| `{"sub":123,"iss":["a"]}` → empty hop, `Extra()` nil, evidence gone | dropped | preserved in `Extra()` |
| dead branches at `:129-133`, `:149-153`, `:178`, `:191-195` — collectively every uncovered statement in the file | unreachable code | gone |

**Test churn is two lines.** In the table runner, `assert.Nil(t, c)` becomes:

```go
require.NotNil(t, c, "a chain must be returned even on error")
assert.True(t, c.Malformed, "error path must set Malformed")
```

and the `act present but not a map` case's `errContains` becomes `"got string"`. Worth adding cases for `act: null`, a malformed `act` past the cap, and `{"sub": 123}`.

**One test gap independent of all the above: the `json:` keys are asserted nowhere.** Mutation-tested on this branch — `chain`→`actors` and `omitted`→`dropped_count,omitempty` each keep `go test ./audit/` green, and the second silently defeats the "truncation is never silent" promise in the `DelegationChain` doc comment. The slog keys *are* well pinned by `TestAuditEventLogTo_Delegation`; it is `TestAuditEventJSON_RoundTrip` that is tautological, since it marshals and unmarshals the same type. Suggested addition:

```go
var wire map[string]any
require.NoError(t, json.Unmarshal(data, &wire))
d := wire["delegation"].(map[string]any)
_, isArray := d["chain"].([]any)
assert.True(t, isArray, `"chain" must be a JSON array, never null`)
assert.Contains(t, d, "omitted", "omitted must be present even when zero")
```

Also worth reconciling with §1: `LogTo` emits `delegation.hops` as an object keyed `hop_0…hop_N`, while `json.Marshal` emits `delegation.chain` as an array. Two shapes for one datum, and ordinal-keyed objects cannot be iterated or filtered in most log query languages.

**Verification.** I applied all four suggestions below plus the two test edits locally on `585e418`: `task lint` clean, `go test -race ./audit/` green, coverage 91.3% → **98.7%** (100% on `ParseDelegationChain`, `countRemainingHops`, and `extraClaims` — the dead branches *were* the uncovered statements), and `delegation.go` gets **27 lines shorter**. Field names are yours to settle per §1; I kept toolhive's `Malformed` spelling on the assumption that convergence is the goal.

Nothing here is blocking on my account once the schema question has an owner — but I would not merge a second incompatible schema without that answer.

audit/delegation.go
```diff
@@ -0,0 +1,227 @@
+// SPDX-FileCopyrightText: Copyright 2025 Stacklok, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package audit
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+)
```
jhrozek, line 9: `encoding/json` becomes unused once the type switch goes: its `json.Number` arm was identical to `default`, and `json.Number` only ever appears when the caller decodes with `UseNumber` anyway.

```suggestion
import (
	"errors"
	"fmt"
)
```

JAORMX:
Thanks for the thorough writeup — this was exactly the right moment to catch it. Direction first: **let's make toolhive-core canonical** and fix toolhive to match. The fact that #6046 already merged shouldn't weigh on what the shared library does; and since stacklok/toolhive#6082 means the token-exchange grant currently has no production traffic, the toolhive wire contract is still cheap to change.

Since your comment I did a deeper survey of the standards and production landscape (RFC 8693/9413, the transaction-tokens / identity-chaining / actor-chain drafts, NIST SP 800-53 AU-family, OWASP ASVS 5.0 V16, MITRE ATT&CK, CoSAI Agentic IAM v1.0, and the audit-log shapes of CloudTrail, GCP, Kubernetes, Entra, Envoy XFCC, Elasticsearch). The PR is updated (585e418 → 4aad6c9) with the result. Point by point:

## 1. Malformed input: you were right, and it's stronger than the audit-sink argument

Core now does **record-and-flag**: `ParseDelegationChain` never fails, sets `malformed: true` plus a low-cardinality `malformedReason` enum (`act_not_object`, `nested_act_not_object`, `iss_not_string`, `sub_not_string`), and keeps every hop parsed before the violation. Sentinel errors are gone. The evidence stack, because this should be a settled question rather than a taste call:

- **MITRE ATT&CK T1562.006 (Indicator Blocking):** if malformed `act` suppresses or degrades the record, anyone who can influence token content gets a remote, unprivileged primitive against the audit trail. This is the decisive argument.
- **CWE-223 / CWE-778** make omission the weakness; **NIST AU-3(f)** requires records to identify the subjects with no "only if parseable" qualifier; UK DPA 2018 s.62 requires identity "so far as possible" — best-effort attribution, record still mandatory.
- **RFC 8693 §4.1's MUST** draws the line for us: only top-level claims + the current actor may inform access control; prior actors are "informational only". A malformed chain therefore *cannot* legitimately fail authorization — being recorded is its only purpose.
- **RFC 9413** (the anti-Postel document) forbids *silent* tolerance, not recording: a loud flag satisfies it; silent dropping violates it twice over ("hiding the consequences of protocol variations… can conceal bugs"). The strictness it asks for belongs at the *sender* — i.e. stacklok/toolhive#6093 should be fixed by rejecting at mint (`invalid_grant`, consistent with the adjacent depth check). Tolerant reader, strict writer; and with that split, `malformed: true` in our own logs becomes a high-signal alert that an issuer is buggy.
- Precedent: syslog RFC 3164 §4.3.3 (the canonical record-and-flag algorithm), CloudTrail's first-class `userIdentity.type: "Unknown"`, ECS `event.outcome: unknown`, OTel's `otel.metric.overflow` sentinel.

The reason field is an enum rather than free text per OTel `error.type` guidance (low cardinality → safe as a metric/alert dimension). That's also my suggestion for your `Malformed` bool during migration: keep the bool, add the reason.

## 2. act_claims: staying with never-serialize, now with citations

NIST AU-3(1)/(3) (limit audit-record content to what's explicitly needed; limit PII to risk-assessed elements), OTel's header-capture rule ("instrumentations SHOULD require explicit configuration of which headers are captured" — JWT claims and HTTP headers are the same threat class: unbounded attacker-influenceable string maps), and the K8s `Metadata` audit level as the production allowlist precedent. One more you'll enjoy: attacker-controlled claim *names* emitted as dynamic log fields can exhaust Elasticsearch's default 1000-field mapping limit and fail ingestion for the whole index — a second indicator-blocking primitive that `act_claims` passthrough exposes and never-serialize closes.

Promoted `iss` stays: OIDC Core §5.7 makes (iss, sub) the only MUST-level stable identifier, and draft-mw-oauth-actor-chain (the current actor-chain standardization vehicle) encodes hops as exactly `{iss, sub}`.

## 3. Your findings checklist — all landed in 4aad6c9

- **`"chain": []` vs `null`:** always allocated now; pinned by a test asserting the emitted bytes. Your prediction was right that our old hard-error path made it unreachable — the flag path makes it reachable, so it's pinned.
- **Emitted-JSON key assertions:** added — tests assert the key sets of the `delegation` object and hops from marshaled bytes, so a tag rename cannot pass green.
- **Non-string `sub`:** flags `sub_not_string` and retains the raw value in the in-memory extra (never serialized) — no more anonymous-actor-with-no-trace. Same treatment for `iss`.
- **Named map types:** reflection fallback accepts `jwt.MapClaims`-shaped input at any nesting level.
- **Cap boundaries:** kept (at-cap, cap+1, depth-1000 exact counts), plus malformed-past-cap now matches your #6096 semantics: flags `malformed`, does not count the unreadable node as omitted.
- **Emitters:** core has exactly one (`LogTo`). Related divergence fixed on our side: slog output now marshals the same shape as the event JSON (single `slog.Any`, `chain` array — the `hop_N` groups are gone), so one parser serves both wire forms. Worth noting for migration: toolhive currently spells the same field three ways (`delegationChain` JSON tag, `delegation_chain` slog key, `delegation_chain` on `PrincipalInfo` vs `delegationChain` in `Identity.MarshalJSON`) — collapsing those to one spelling is part of the win.
- **Depth 10 vs 16:** kept 16 with the distinction documented on the constant, exactly as you suggested — parse-time defensive ceiling deliberately above any mint cap, so `truncated: true` remains a "foreign token" signal for a consumer passing its mint cap. FWIW 10 as an *enforcement* default now has a citable source: draft-mw-spice-actor-chain §21.9 (RECOMMENDED default 10).

## 4. Ordering: outermost-first, now deliberate rather than coincidental

We already agreed, but the survey turned up a real hazard worth documenting (and I did, on the type): every *append-style* flat chain out there (Envoy XFCC, GCP `delegates[]`, SPICE `ach`) runs the **opposite** direction (original-first). Current-actor-first is still right for this type: it matches RFC 8693's nested read order, and — decisive for an audit record with truncation — it keeps `Chain[0]` stable under truncation, since truncation drops *inner* hops. Original-first ordering would let index 0 silently hold a mid-chain actor on truncated chains. (Amusingly, #6035's acceptance criteria ask for chronological original-first while its own proposal says outermost-first; the RFC's order won.)

## 5. Migration path for toolhive

Happy to drive this. It's smaller than the table suggests because #6096 already moved your parser to value-return + `Malformed` — semantically we've now converged on your model. Remaining deltas: renames (`delegationChain`→`delegation`, `actors`→`chain`, `dropped_count`→`omitted` incl. dropping its `omitempty`), promote `iss` out of `act_claims`, drop `act_claims` serialization entirely (which also makes the `internalClaims`/`tsid` per-hop filter moot — the leak becomes structurally impossible), adopt `malformedReason`, unify the three wire spellings, emit the chain via marshaling in `LogTo`, and keep passing `maxDepth=10` at the call sites. Plus fixing #6093 at mint per point 1. `docs/middleware.md` examples get rewritten once, before any consumer exists (#6082).

audit/delegation.go
JAORMX: Right on all counts — and my 4aad6c9 rewrite had the same oversell ("bounds the work done"). Adopted your wording adapted to the current text in c7d4de0: parse ceiling, bounds the size of the retained chain and emitted record, explicitly does not bound parsing work. The ceiling-vs-mint-cap distinction was already documented on the constant.

audit/delegation.go
JAORMX: The sub/act conflation was real — fixed in c7d4de0 with your phrasing ("the party that presented the token"; "carried by the event's own Subjects map; the chain never repeats it").

`Malformed` landed in 4aad6c9 with one addition over your suggestion: a low-cardinality `malformedReason` enum alongside the bool (`act_not_object` / `nested_act_not_object` / `iss_not_string` / `sub_not_string`), so the *kind* of violation is queryable in the record and safe as a metric/alert dimension (per OTel `error.type` guidance — enum, not free text). All three status fields are always-emitted, no omitempty.

audit/delegation.go
JAORMX: We crossed mid-air — 4aad6c9 landed essentially this design (never-nil always-allocated chain, `Malformed` flag, null terminates the chain, hops-before-the-violation kept, the four dead branches gone) with three deltas worth your eyes:

- **No error return at all.** I went one step further than the `(chain, error)` contract: the parse cannot fail, and the violation is recorded as `malformedReason` on the chain itself. Rationale: (a) the reason lands **in the record**, where investigators and alert rules can see it, rather than in an error the caller must remember to log; (b) `(chain, error)` invites the exact anti-pattern this feature exists to prevent — an `if err != nil { return }` at a call site silently drops the record (toolhive's `claimsToIdentity` would face precisely that temptation); (c) with flag+reason on the struct there's nothing left for `errors.Is` to branch on. The `%T` fidelity is the one real loss, traded for the reason being a bounded enum. Happy to restore the error alongside the non-nil chain if you feel strongly — that contract works too, it's just easier to hold wrong.
- **Named map types** (`jwt.MapClaims{"act": jwt.MapClaims{...}}` at any level) parse via a small reflection fallback instead of being misread as non-objects — your own #6046 review finding.
- **A malformed tail past the cap** sets `Malformed` (not counted in `Omitted`, since a non-object isn't an actor) rather than passing silently — a truncated chain with an unreadable tail seems doubly worth flagging.

audit/delegation.go
JAORMX: Implemented exactly this way in 4aad6c9, with one addition: the parse also flags the chain (`iss_not_string` / `sub_not_string`), so the wire record itself shows that an identifier was present but unusable — while the raw value stays inspectable in the in-memory extra and never reaches the wire.

audit/delegation.go
JAORMX: Moot as of 4aad6c9 — the type switch is gone entirely (any non-map input takes the malformed path uniformly), and with the sentinel errors removed the file now imports only `reflect`.

audit/event.go
```diff
@@ -185,6 +206,17 @@ func (e *AuditEvent) LogTo(ctx context.Context, logger *slog.Logger, level slog.
...
+	// key is emitted). The chain is emitted via its JSON marshaling so the
+	// slog and JSON wire shapes of the same event are structurally identical:
+	// a "chain" array of {iss,sub} hops plus truncated/omitted/malformed.
+	// Extra per-hop claims are never serialized (PII guard).
+	if !e.DelegationChain.IsZero() {
+		attrs = append(attrs, slog.Any("delegation", e.DelegationChain))
+	}
+
 	// Log with the specified level
 	logger.LogAttrs(ctx, level, "audit_event", attrs...)
```
jhrozek, line 221: 🔴 **This leaks the extra claims under any non-JSON handler.** `slog.Any` stores the pointer as `KindAny`; `JSONHandler` marshals it with `encoding/json`, which skips the unexported field, but `TextHandler` formats it with `%+v`, which does not. Verified at `c7d4de0`:

```
delegation="&{Chain:[{Issuer:outer Subject:current
  extra:map[email:secret@example.com tsid:internal-session-9f3a]}]
  Truncated:false Omitted:0 Malformed:false MalformedReason:}"
```

JSON leaks: `false`. Text leaks: `true`, including `tsid` — the exact internal-claim class cited as the reason never-serialize beats filtering. It is one option call away, not exotic: the sibling package exposes `logging.WithFormat(logging.FormatText)` (`logging/logging.go:96`), and text mode is what people run locally, against real IdPs. The `hop_N` code this replaced was safe here precisely because it built the attrs by hand, so `extra` never reached a handler; text output also loses its structure now, becoming one Go-syntax blob.

`LogValue` fixes both — slog resolves `LogValuer` *before* the handler formats anything, so the call site stays as-is and the JSON shape is unchanged. Note the `omitempty` mirroring: without it an absent `sub` emits `""` in slog but is omitted in the event JSON, which silently re-breaks the parity this commit set out to establish (my first draft had that bug; a byte-comparison across six fixtures caught it).

Verified locally on `c7d4de0`: `task lint` clean, `go test -race ./audit/` green, and `delegation` byte-identical between `json.Marshal(event)` and `JSONHandler` output across normal / empty-hop / iss-only / non-string-sub / non-object / 20-deep-truncated fixtures.

```suggestion
	// Add delegation chain if present and non-zero. The IsZero gate plus the
	// field's omitempty tag together guarantee that, when no chain is set,
	// the log output is identical to before this field existed (no "delegation"
	// key is emitted). The chain resolves through [DelegationChain.LogValue],
	// which mirrors its JSON shape, so the slog and JSON wire forms of the same
	// event are structurally identical.
	if !e.DelegationChain.IsZero() {
		attrs = append(attrs, slog.Any("delegation", e.DelegationChain))
	}

	// Log with the specified level
	logger.LogAttrs(ctx, level, "audit_event", attrs...)
}

// LogValue implements [log/slog.LogValuer] so that every handler — not just
// JSONHandler, which happens to skip unexported fields via encoding/json —
// sees only the promoted iss/sub of each hop. Handing the struct to a handler
// directly lets a text or other fmt-based handler render the unexported extra
// claims via %+v, defeating the PII guard. The emitted shape mirrors the
// struct's JSON tags so slog and JSON output stay structurally identical.
func (c *DelegationChain) LogValue() slog.Value {
	hops := make([]map[string]string, 0, len(c.Chain))
	for _, hop := range c.Chain {
		// Mirror the omitempty tags on DelegatedActor so the shapes stay
		// identical for hops with an absent iss or sub.
		logged := make(map[string]string, 2)
		if hop.Issuer != "" {
			logged[hopClaimIss] = hop.Issuer
		}
		if hop.Subject != "" {
			logged[hopClaimSub] = hop.Subject
		}
		hops = append(hops, logged)
	}
	attrs := []slog.Attr{
		slog.Any("chain", hops),
		slog.Bool("truncated", c.Truncated),
		slog.Int("omitted", c.Omitted),
		slog.Bool("malformed", c.Malformed),
	}
	if c.MalformedReason != "" {
		attrs = append(attrs, slog.String("malformedReason", string(c.MalformedReason)))
	}
	return slog.GroupValue(attrs...)
}
```

audit/delegation_test.go
```diff
@@ -0,0 +1,789 @@
...
+	logger := slog.New(slog.NewJSONHandler(&buf, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	event.LogTo(context.Background(), logger, LevelAudit)
+	var fromSlog map[string]any
+	require.NoError(t, json.Unmarshal(buf.Bytes(), &fromSlog))
+
+	require.Contains(t, fromJSON, "delegation")
+	require.Contains(t, fromSlog, "delegation")
+	assert.Equal(t, fromJSON["delegation"], fromSlog["delegation"],
+		"slog and JSON delegation shapes must be identical")
+}
```
jhrozek, line 724: This test is the right idea but cannot catch the leak above: it only exercises `JSONHandler`, which is the one handler that is safe by accident. The fixture also gives every hop both an `iss` and a `sub`, so it cannot catch an `omitempty` parity break either.

Extending it to a text handler plus a hop with no `sub` closes both. Mutation-checked: with `LogValue` renamed away (i.e. today's behaviour) this fails on all three new assertions; with the fix applied it passes.

```suggestion
func TestAuditEventLogTo_SlogMatchesJSON(t *testing.T) {
	t.Parallel()

	// The outer hop carries a PII extra claim; the inner hop has no sub. The
	// extra must reach no handler at all, and an absent sub must be omitted
	// rather than emitted as "" so both wire shapes stay identical.
	chain := ParseDelegationChain(map[string]any{
		hopClaimIss:     testIss1,
		hopClaimSub:     testSub1,
		extraClaimEmail: testAliceEmail,
		hopClaimAct: map[string]any{
			hopClaimIss: testHopInner,
			hopClaimAct: "broken", // exercises malformed + malformedReason too
		},
	}, 0)

	event := NewAuditEvent("t", EventSource{}, OutcomeSuccess, map[string]string{}, "svc")
	event.WithDelegationChain(chain)

	data, err := json.Marshal(event)
	require.NoError(t, err)
	var fromJSON map[string]any
	require.NoError(t, json.Unmarshal(data, &fromJSON))
	require.Contains(t, fromJSON, "delegation")

	var buf bytes.Buffer
	logger := slog.New(slog.NewJSONHandler(&buf, &slog.HandlerOptions{Level: slog.LevelDebug}))
	event.LogTo(context.Background(), logger, LevelAudit)
	var fromSlog map[string]any
	require.NoError(t, json.Unmarshal(buf.Bytes(), &fromSlog))
	require.Contains(t, fromSlog, "delegation")
	assert.Equal(t, fromJSON["delegation"], fromSlog["delegation"],
		"slog and JSON delegation shapes must be identical")

	// Every handler, not only JSONHandler: a fmt-based handler must not be able
	// to render the unexported extra claims, and must keep the chain structured.
	var text bytes.Buffer
	event.LogTo(context.Background(),
		slog.New(slog.NewTextHandler(&text, &slog.HandlerOptions{Level: slog.LevelDebug})), LevelAudit)
	assert.NotContains(t, text.String(), testAliceEmail,
		"extra claim values must not reach a text handler")
	assert.NotContains(t, text.String(), extraClaimEmail,
		"extra claim keys must not reach a text handler")
	assert.Contains(t, text.String(), "delegation.chain=",
		"the chain must stay structured in text output")
}
```

audit/doc.go
```diff
@@ -26,16 +26,50 @@ Each audit event records:
...
+# Delegation Chains
+
+When an actor acted on behalf of another actor (RFC 8693 `act` claim), attach
+a [DelegationChain] to the event via [AuditEvent.WithDelegationChain]. The
+chain is additive to the flat [AuditEvent.Subjects] map; Subjects remains the
+backward-compatible identity map and is not modified by setting a chain.
+
+Parse raw `act` claims with [ParseDelegationChain], which applies a defensive
+depth cap ([DefaultMaxDelegationDepth], overridable via maxDepth) to bound
+work on attacker-influenceable input, and reports explicit truncation
```
jhrozek: c7d4de0 fixed this wording on the constant itself but not here, so the package doc now contradicts it — this still claims the cap bounds *work*, while `DefaultMaxDelegationDepth` correctly says it does not.

```suggestion
Parse raw `act` claims with [ParseDelegationChain], which applies a defensive
depth cap ([DefaultMaxDelegationDepth], overridable via maxDepth) to bound the
size of the retained chain and of the emitted record — not parsing work, since
the caller has already decoded the claim — and reports explicit truncation
```

audit/delegation.go
```diff
@@ -0,0 +1,321 @@
...
+	Malformed       bool             `json:"malformed"`
+	MalformedReason MalformedReason  `json:"malformedReason,omitempty"`
+}
+
+// IsEmpty reports whether the chain is nil or contains no hops. Note that a
+// chain with no hops may still carry audit-relevant state (Malformed); use
+// [DelegationChain.IsZero] to decide whether a chain is worth recording.
+func (c *DelegationChain) IsEmpty() bool {
+	return c == nil || len(c.Chain) == 0
+}
```
jhrozek: Nit, no suggestion since the test needs a coordinated edit: `IsEmpty` is now dead inside the package and reads as a footgun next to `IsZero` — a hopless-but-malformed chain *is* empty, yet is exactly the case that must still be recorded, and the doc comment has to warn about that. It was added in 585e418 and has never shipped, so deleting it and keeping only `IsZero` costs nothing but the assertions in `TestDelegationChain_IsEmptyIsZero`. One predicate, no way to pick the wrong one.

jhrozek (changes requested):
## Re-review of 4aad6c9 + c7d4de0 — nearly all of it lands, one new regression

I re-checked every finding against the code rather than the summary, and mutation-tested the parts that claimed new coverage. Verified fixed:

| finding | evidence |
|---|---|
| malformed → record-and-flag | `ParseDelegationChain` has no error return, sentinels gone, `MalformedReason` enum, hops before the violation preserved |
| **depth invariance** | `h1→h2→h3→act:"garbage"` now yields a record at *both* `maxDepth=2` (2 hops, `omitted=1`, `malformed`) and `maxDepth=16` (3 hops, `malformed`) — previously a record vs. a hard error |
| inflated `omitted` | `countRemainingHops` returns `(int, bool)`; the non-object is flagged, not counted |
| `"chain": null` | always allocated, pinned from marshaled bytes |
| `json:` keys unasserted | re-ran all three mutations — `chain`→`actors` (7 failures), `omitted`→`dropped_count,omitempty` (fails), `flagMalformed` neutered (fails). Was green on all three before |
| non-string `iss`/`sub` | flags `iss_not_string`/`sub_not_string`, raw value retained in the in-memory extra |
| `act: null` | terminates the chain |
| dead branches / unused import | gone — 100% statement coverage on all eight functions in `delegation.go`, 99.0% package |
| §6 miscitation, actor/subject conflation | corrected, and OIDC Core §5.7 is the right authority for `(iss, sub)` |
| depth-cap rationale (c7d4de0) | now says exactly what it does — parse ceiling on retained size, explicitly not work |

`task lint` clean, `go test -race ./audit/` green at `c7d4de0`.

Two things I'll concede outright. The MITRE T1562.006 framing for record-and-flag is a better argument than the audit-fidelity one I made — "malformed input can suppress the audit trail" is a security property, not a quality preference, and `malformedReason` as a low-cardinality enum makes it alertable in a way a bare bool isn't. And the ordering-rationale note about Envoy XFCC / GCP `delegates[]` running original-first is a hazard I did not flag and should have.

**The one blocker is a regression introduced by the shape unification.** Passing `*DelegationChain` straight to `slog.Any` is PII-safe only because `JSONHandler` happens to route through `encoding/json`, which skips unexported fields. `TextHandler` formats with `%+v` and prints `extra` in full — including the `tsid`-class internal claim that never-serialize exists to make unrepresentable. Verified at `c7d4de0`, details inline. The fix is ~20 lines and keeps both wire shapes identical; I applied it and the suggested test locally before proposing them.

Marking this `REQUEST_CHANGES` only for that one item — everything else in the delegation work is in better shape than when I first read it.

audit/event.go
JAORMX: Confirmed and fixed in eb00bd0 — this was a real regression from the shape unification, and the worst kind: safe on the handler everyone tests, leaking on the handler people run locally against real IdPs. Adopted your `LogValue` verbatim including the omitempty mirroring (placed in delegation.go next to the type rather than event.go, since it's the type's contract); the LogTo comment now names `LogValue` as the mechanism. Thanks for mutation-testing it before proposing — the sub-less-hop parity case is exactly the bug I'd have reintroduced.

audit/delegation_test.go
JAORMX: Adopted verbatim in eb00bd0. Verified both directions locally: with `LogValue` present the three new assertions pass; renaming it away fails all three (email value + key reach the text handler, and the chain degrades to the `%+v` blob).

audit/doc.go
JAORMX: Applied in eb00bd0 — the package doc and the constant now say the same thing.

audit/delegation.go
JAORMX: Agreed — deleted in eb00bd0 along with its test assertions (`TestDelegationChain_IsEmptyIsZero` → `TestDelegationChain_IsZero`). One predicate, and it's the one whose semantics match the recording decision. Never shipped, so free.

audit/delegation.go
```diff
@@ -0,0 +1,348 @@
...
+// OpenID Connect Core §5.7 the (iss, sub) combination is the only stable,
+// unique actor identifier, and RFC 8693 §6 calls for data minimization. Any
+// other claims present on a hop are retained in the in-memory-only extra map
+// (never serialized or logged) to avoid leaking Personally Identifiable
+// Information.
+type DelegatedActor struct {
+	Issuer  string `json:"iss,omitempty"`
+	Subject string `json:"sub,omitempty"`
+	extra   map[string]any
+}
```
jhrozek, line 69: Non-blocking, defense-in-depth. The guard holds for `*DelegationChain`, which is the only shape `LogTo` emits — but `LogValue` has a pointer receiver, so the value type and `DelegatedActor` don't implement `LogValuer`, and a consumer can route around it:

```
slog.Any("d", *chain)         → leak=true   d="{Chain:[{Issuer: Subject:s extra:map[email:secret@example.com]}]…}"
slog.Any("d", chain.Chain[0]) → leak=true
```

`fmt` never consults `LogValuer`, but it *does* consult `Stringer` — so one method on the hop closes both doors, including any future struct that embeds a hop. Verified: `%+v` of a value, a pointer, and a bare hop all render `{iss: sub:s}` with no leak; suite green, lint clean.

`Extra()`'s doc already tells callers not to serialize the map, so the hazard is acknowledged — this just makes it structural rather than advisory, on the one type whose selling point is that the extras *cannot* leak.

```suggestion
type DelegatedActor struct {
	Issuer  string `json:"iss,omitempty"`
	Subject string `json:"sub,omitempty"`
	extra   map[string]any
}

// String implements [fmt.Stringer] so that fmt-based rendering of a hop — or of
// any struct containing hops, such as a dereferenced DelegationChain — cannot
// reach the unexported extra claims through %v/%+v. fmt does not consult
// [log/slog.LogValuer], so the PII guard needs this second, fmt-side door
// closed as well.
func (a DelegatedActor) String() string {
	return "{iss:" + a.Issuer + " sub:" + a.Subject + "}"
}
```

audit/delegation.go
```diff
@@ -0,0 +1,348 @@
...
+
+// LogValue implements [log/slog.LogValuer] so that every handler — not just
+// JSONHandler, which happens to skip unexported fields via encoding/json —
+// sees only the promoted iss/sub of each hop. Handing the struct to a handler
+// directly would let a text or other fmt-based handler render the unexported
+// extra claims via %+v, defeating the PII guard. The emitted shape mirrors the
+// struct's JSON tags (including omitempty on iss/sub and malformedReason) so
+// slog and JSON output stay structurally identical.
+func (c *DelegationChain) LogValue() slog.Value {
+	hops := make([]map[string]string, 0, len(c.Chain))
```
jhrozek, line 154: Also non-blocking: `LogValue` is now the only method on the type without the nil guard `IsZero` has, and slog resolves `LogValuer` on nil pointers. `slog.Any("d", (*DelegationChain)(nil))` currently yields:

```
d="LogValue panicked\n…audit.(*DelegationChain).LogValue (delegation.go:142)…"
```

`slog.Value.Resolve` recovers, so nothing crashes — but the record carries a stack trace where the value should be. With the guard, an empty group is elided and the attribute simply isn't emitted, which is the honest rendering of a nil chain. Verified alongside the `String()` suggestion: suite green, lint clean.

```suggestion
func (c *DelegationChain) LogValue() slog.Value {
	if c == nil {
		return slog.GroupValue()
	}
	hops := make([]map[string]string, 0, len(c.Chain))
```

jhrozek (dismissed):
## ✅ Everything I raised is resolved — approving

Re-verified `eb00bd0` by exercising the code rather than reading the diff:

| item | verification |
|---|---|
| `LogValue` PII guard | text handler `leak=false`; mutation — rename `LogValue` away and the suite **fails** on all three new assertions |
| slog/JSON parity, incl. `omitempty` | `delegation` object byte-identical between `json.Marshal(event)` and `JSONHandler` across six fixtures: normal, empty hop, iss-only, non-string sub, non-object, 20-deep truncated |
| text-handler test | adopted verbatim; it is what the mutation trips |
| `doc.go` wording | package doc and the constant now say the same thing |
| `IsEmpty` deleted | no references left; test renamed to `TestDelegationChain_IsZero` |

`task lint` clean, `go test -race ./audit/` green, coverage **99.1%** — the sole uncovered block is the pre-existing `e.Data` path in `LogTo`, untouched here.

Putting `LogValue` on the type in `delegation.go` rather than next to `LogTo` was the right call, and I'd have argued for it if you hadn't done it — it's the type's contract, not the event's.

Two notes on the earlier rounds, since you asked directly. On **`(chain, error)` vs. no error at all**: your (b) is the argument that settles it — `if err != nil { return }` at a call site is exactly the silent-drop this feature exists to prevent, and a contract that's easy to hold wrong in an audit path is a worse contract. With flag+reason on the struct there is genuinely nothing left to branch on. The `%T` fidelity loss is real but cheap next to that, and `malformedReason` recovers most of it as a bounded enum. Keep it as is. On **flagging a malformed tail past the cap**: agreed, and it's the fix that restored depth invariance — the same token now produces a record with the same `malformed` signal at any `maxDepth`, where before it was a record at 2 and a hard error at 16.

One residual below, explicitly non-blocking: the guard is airtight for the shape `LogTo` passes, and porous for three shapes a *consumer* can construct. Fine to land as-is and follow up, or take the two suggestions.

audit/delegation.go
JAORMX: Took it now rather than as a follow-up — 4eeaad5, `String()` adopted verbatim, plus a test rendering a bare hop, a dereferenced chain, and the pointer through `%v`/`%+v` and asserting neither the extra key nor value appears (and that the subject still does). Structural beats advisory, agreed.

audit/delegation.go
JAORMX: Also in 4eeaad5, with a test that a nil chain through `slog.Any` produces no "panicked" rendering. An elided empty group is indeed the honest output for a nil chain.

JAORMX:
@jhrozek the 4eeaad5 push auto-dismissed your approval — it contains only your two non-blocking suggestions applied verbatim (`String()` on the hop + the `LogValue` nil guard) plus tests pinning both. CI is green across all five checks and every thread is resolved. Mind re-approving so this can land?

jhrozek (approved):
## ✅ Re-approving `4eeaad5`

Both hardening suggestions landed, and you pinned each one rather than just applying it — `TestDelegatedActor_FmtCannotLeakExtra` and `TestDelegationChain_LogValueNilSafe`. Mutation-checked all three guards; each fails the suite when removed:

| guard removed | result |
|---|---|
| `DelegatedActor.String()` | ❌ FAIL |
| `LogValue` nil guard | ❌ FAIL |
| `LogValue` itself (the earlier fix) | ❌ FAIL |

Independent behavioural check at this SHA, not relying on the new tests:

```
fmt  %+v hop / %v value / %+v pointer   leak=false   {iss:i sub:s}
slog pointer / value / bare hop         leak=false
slog nil pointer                        no attribute emitted
slog delegation == json delegation      true
```

`task lint` clean, `go test -race ./audit/` green, coverage 99.1% with `String` and `LogValue` at 100% — the only uncovered block remains the pre-existing `e.Data` path in `LogTo`, untouched by this PR.

Nothing outstanding from my side. Three rounds' worth of blocking items — the schema divergence with the merged toolhive#6046, the `(nil, err)` contract that let a malformed `act` erase the chain from the record, and the `TextHandler` PII leak from the shape unification — are all resolved, each with a test that fails if the fix is reverted.

PR #230, opened 2026-08-12
Add container/signer for cosign key-based OCI signing

samuv:
## Summary

Sigstore *verification* already lives here, in `container/verifier`. Signing lived in toolhive (`pkg/skills/signer`), so the two halves of one trust model sat in different repositories — with a visible cost: the signer's round-trip test could not use the real verifier, and hand-rolled a ~20-line equivalent instead, carrying the comment *"used directly here until the core release this stack's verifier PR consumes"*.

This ports the package to `container/signer`, next to its counterpart. It was written to be moved and that held up:

- it imported **nothing** from toolhive — stdlib, go-containerregistry, sigstore-go, sigstore, protobuf-specs, go-digest
- every one of those is already a **direct** dependency of this module, so `go.mod` is untouched
- no import cycle: `container/verifier` does not reference the signer

What it does: signs an OCI artifact following the cosign key-pair convention — a simple-signing payload binding the manifest digest is signed with the user's key, attached to the registry as a cosign signature manifest at the `sha256-<hex>.sig` tag, and returned as a serialized Sigstore bundle for durable storage and offline re-verification. Keyless (OIDC) signing is not implemented; `ErrKeyRequired` is returned instead.

Three small generalizations, since this is now a library rather than skills-specific code:

- package doc says "OCI artifacts" rather than "skill OCI artifacts", and cross-references `container/verifier`
- `ErrKeyRequired` no longer names a `--key` CLI flag that does not exist at this layer; its doc comment tells CLI callers to wrap it with their own flag name
- two comments referring to "a `--key` flag" now say "caller input (typically a CLI flag)"

## Test plan

- [x] Unit tests (`task test`)
- [x] Linting (`task lint-fix`)

All nine existing tests ported and passing, plus the whole `container/...` tree green.

The round-trip test is now meaningfully stronger. It previously verified signer output against a locally reconstructed verifier; it now calls `verifier.VerifyBundleOfflineWithKey`, so sign → verify is proven against the entry point consumers actually use, in one repository:

```go
func verifyKeyBundle(t *testing.T, raw, pubPEM []byte, payloadDigest []byte) error {
	t.Helper()
	_, err := verifier.VerifyBundleOfflineWithKey(
		raw,
		verifier.DigestAlgorithmSHA256+":"+hex.EncodeToString(payloadDigest),
		pubPEM,
	)
	return err
}
```

Coverage retained: round trip against a live in-process registry, wrong-key rejection, encrypted-key handling via `COSIGN_PASSWORD`, missing-key error, digest normalization, payload/digest binding, keypair metadata, and `resolveKeyPath` path vetting.

## Notes for reviewers

**Nothing is removed from toolhive in this PR.** Its `pkg/skills/signer` stays until this lands and ships in a release, because an in-flight toolhive PR (stacklok/toolhive#6139) still modifies that package. Deleting it now would conflict. The toolhive-side removal is a follow-up gated on both.

`resolveKeyPath` is worth a look — it is the sanitization barrier for a user-supplied key path (absolute, cleaned, symlinks resolved, regular-file checked) and carries the `//nolint:gosec` that depends on it. It was written to satisfy CodeQL's path-injection analysis in toolhive; flag it if this repo wants a different approach.


samuv:
Pushed a second commit: the ported package could not read keys from `cosign generate-key-pair`, so the interop claim in its docs did not hold.

Found by signing a real artifact with `thv skill push --key` and verifying it with the stock cosign CLI. `cosign` seals PKCS#8 with scrypt + nacl/secretbox; `cryptoutils` writes PKCS#8 standard encryption. Both use the `ENCRYPTED SIGSTORE PRIVATE KEY` label, and only the second was handled — so the existing fixtures, built with `cryptoutils`, looked like cosign keys but were not.

Both formats are accepted now, with a wrong password reported as such instead of as an ASN.1 parse error. New tests build fixtures with `encrypted.Encrypt`, the same call cosign makes.

The same fix is in stacklok/toolhive#6139, which ships first; this keeps the two in step so the eventual move does not regress it.

samuv:
Third commit, and this one touches a security workflow so it deserves explicit attention.

Adding `container/signer` made `GO-2026-5932` reachable and failed the vulnerability check:

```
Vulnerability #1: GO-2026-5932
    The golang.org/x/crypto/openpgp package is unmaintained, unsafe by design
    Fixed in: N/A
      #1: container/signer/key.go:18:2: signer.init calls sign.init, which eventually calls armor.init
```

It is genuinely introduced by this PR — `main` passes. The cause is that linking `sigstore-go/pkg/sign` pulls openpgp in through package init for backward-compatible OpenPGP support. There is no fixed version and none is planned; the package is deprecated by design.

**stacklok/toolhive already excludes this exact advisory** with the same justification, because it consumes the same sigstore packages. Core had no exclusion mechanism at all — the action's result was the whole gate — so I added one modelled on toolhive's: parse the JSON findings, ignore only documented IDs, and **fail on anything else**. Net effect is stricter than before for every other advisory, since the job now explicitly errors on unexcluded findings rather than relying on the action's exit code alone.

Worth a reviewer's judgement on whether you want that mechanism here, or would rather solve it another way. The alternative I can see is not linking `sigstore-go/pkg/sign`, which would mean not having a signer.

samuv:
Tightened the exclusion's justification after digging into your question about swapping in ProtonMail's library.

Short answer: we cannot, and `gopenpgp` would be the wrong one anyway.

```
container/signer → sigstore-go/pkg/sign → rekor/pkg/pki → rekor/pkg/pki/pgp → x/crypto/openpgp
```

Nothing here imports openpgp — `rekor/pkg/pki` is a pluggable format registry, so importing it links every format including PGP. There is no call site to rewrite. And `ProtonMail/gopenpgp/v3` is a high-level wrapper with a different API; the API-compatible fork is `ProtonMail/go-crypto/openpgp`. Even that cannot be forced with a `replace`, since it would have to swap the whole `golang.org/x/crypto` module while go-crypto only carries the openpgp subtree.

**But the question surfaced something useful: rekor has already done it.** [sigstore/rekor#2883](https://github.com/sigstore/rekor/pull/2883) migrated to `ProtonMail/go-crypto/openpgp` on 2026-07-15 — after v1.5.3 shipped on 2026-07-02, so it is on `main` and unreleased.

So the exclusion's expiry is now a real condition rather than "when sigstore drops the dependency": bump rekor past v1.5.3 once a release contains the migration. The comment says that, plus the import chain and the fact that every govulncheck trace is package-init reachability rather than a call — which is the part that actually makes the exclusion defensible.

Tracked in stacklok/toolhive-core#231, and stacklok/toolhive#6287 for the same exclusion there.

rdimitrov (dismissed):

rdimitrov (approved):

PR #235, opened 2026-08-13
Add Fulcio/Rekor keyless signing to SignOCI

samuv:
## Summary

`container/signer.Default.SignOCI` could only sign with a file-based cosign key, returning `ErrKeyRequired` when none was given. That left the certificate-bearing attach and retrieval paths — fixed in #234 (PR C1) — without a producer: nothing in-tree could actually mint a keyless signature, so the layout could only be exercised with hand-built fixtures.

This adds the real keyless path. Part of the keyless push-signing work tracked as stacklok/toolhive#6307.

- **`Options.IdentityToken` selects the keyless flow.** A single-use ephemeral key pair is minted, Fulcio issues a short-lived certificate binding it to the token's identity, and the signature is submitted to Rekor. Acquiring the token (ambient CI credentials, browser flow, device flow) stays entirely the caller's problem — this package never performs OAuth and only forwards the token it is handed.
- **A Rekor entry is mandatory, not optional.** `container/verifier`'s own policy for keyless bundles requires `WithSignedCertificateTimestamps(1)`, `WithTransparencyLog(1)` and `WithObserverTimestamps(1)`, so a bundle with no transparency-log entry would not verify under the policy this repo itself applies. Rekor **v1** specifically: v2 issues proof-only entries and needs a separate timestamp authority, and the cosign bundle annotation has no field for a v2 inclusion proof.
- **`Options.FulcioURL` / `Options.RekorURL`** override the public-good endpoints, defaulted at signing time so a zero-valued `Options` targets public-good.
- **Both `Key` and `IdentityToken` set is an error**, not a precedence rule. The two produce trust material verifiers check in materially different ways (a bare public key versus a Fulcio identity plus the transparency log), so resolving it silently would attach a signature the caller cannot verify the way it expects. New sentinel `ErrAmbiguousSigningMethod`.
- **`ErrKeyRequired`'s message now names both options.** Phrased in terms of the `Options` fields, since this package does not know CLI flag names.

`attachCosignSignature` and `container/verifier` are untouched — C1 already handles cert-bearing bundles correctly, and the keyless path calls it with exactly the same arguments as the key path.

### Where defaulting happens, and why the URLs are hardcoded

Both documented on `Options` and the constants. Defaulting is applied inside `keylessBundleOptions` at signing time. The endpoints are hardcoded rather than derived from a trusted root because reading them means fetching a root (a TUF round trip, or the point-in-time snapshot embedded in `container/verifier`) purely to learn a URL, and the URI is only reachable by type-asserting `root.CertificateAuthority` — a single-method `Verify` interface — to the concrete `*root.FulcioCertificateAuthority`. More importantly the trusted root is *verification* material: keying signing endpoints off it would let a trust-root refresh silently redirect where signing requests, and the identity token in them, are sent.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Changes

| File | Change |
| --- | --- |
| `container/signer/signer.go` | `Options.IdentityToken` / `FulcioURL` / `RekorURL`; `signingMaterial` + `keylessBundleOptions`; `ErrAmbiguousSigningMethod`; updated `ErrKeyRequired` |
| `container/signer/keyless_test.go` | In-process Fulcio + CT log + Rekor v1 over `httptest`, doubling as the verification trust root; keyless tests |
| `go.mod` | Test-only imports promoted from indirect to direct (`certificate-transparency-go`, `rekor`, `go-openapi/runtime`, `swag/conv`, `json-canonicalization`). No new modules, `go.sum` unchanged |

## Test plan

- [x] Unit tests pass (`task test`, full suite, race detector)
- [x] Lint passes (`task lint`, 0 issues)
- [x] `go build ./container/...` and `go vet ./container/...`
- [x] Existing key-signed tests pass **unmodified** (`TestSignOCIRoundTrip`, `TestSignOCIEncryptedKey`, `TestSignOCIAcceptsCosignCLIKeyFormat`, the C1 attach tests, …)

**This PR is network-free.** Real Sigstore-staging E2E coverage is separate follow-on work, not this PR.

### How the Fulcio/Rekor interaction is tested without the network

`testSigstore` is a complete Sigstore deployment running in-process behind `httptest` servers that speak the actual HTTP contracts sigstore-go's clients drive — `POST /api/v2/signingCert` and `POST /api/v1/log/entries`. Servers, not stubs: mocking sigstore-go's `CertificateProvider` / `RekorClient` interfaces would have left the part most likely to break — that `SignOCI` wires the ephemeral key, identity token and endpoint URLs into real requests — untested.

The material it issues is cryptographically genuine, which is what lets the same object also implement `root.TrustedMaterial` and serve as the verification trust root:

- **Fulcio** verifies the caller's proof of possession before issuing. That check only passes if `SignOCI` signed the token's subject with the very key it asked to have certified.
- **Certificates carry a real embedded SCT.** The certificate is issued twice on purpose: an embedded SCT signs the *pre*-certificate's TBS, so the pre-certificate must exist before the SCT can be computed, and the final certificate is reissued from the same template with the SCT appended.
- **Rekor unmarshals the proposed entry through Rekor's own type registry**, which cryptographically validates the submitted signature against the entry's public key and artifact hash exactly as the real log does, then canonicalizes it and signs a real `SignedEntryTimestamp` and single-leaf checkpoint.

Coverage:

| Test | What it pins |
| --- | --- |
| `TestSignOCIKeylessRoundTrip` | Attach → `verifier.RetrieveBundles` → `HasCertificate()` is true |
| `TestSignOCIKeylessBundleVerifies` | Passes `verifier.VerifyBundle` with `DefaultVerifierOptions()` against the test trust root; identity read back and re-pinned; a different identity is rejected |
| `TestSignOCIKeylessRejectedByUnrelatedTrustRoot` | Guards the above against passing for the wrong reason — a second, independent deployment rejects the bundle |
| `TestSignOCIKeylessAttachesCertificateAndTlogAnnotations` | Certificate and Rekor bundle annotations, and the SAN/issuer in the cert |
| `TestSignOCIKeylessReSignAppendsForNewIdentity` | C1's identity dedupe against *real* Fulcio-issued certs: same identity dedupes, different identity appends |
| `TestSignOCIKeylessSurfacesFulcioFailure` | A rejected token fails signing rather than falling back to another layout |
| `TestSignOCIRejectsAmbiguousSigningMethod`, `TestErrKeyRequiredNamesBothSigningMethods` | The two error paths |

I mutation-tested the two assertions that could quietly stop covering anything, and confirmed both fail when the fake stops doing its job: dropping the SCT extension yields `only able to verify 0 SCT entries; unable to meet threshold of 1`, and corrupting the `SignedEntryTimestamp` yields `not enough verified log entries from transparency log: 0 < 1`.

## Special notes for reviewers

- The `go.mod` churn is only indirect → direct promotion for test-only imports; `go.sum` is unchanged.
- `sign.BundleOptions.TrustedRoot` is deliberately left unset, so `sign.Bundle` does not verify the bundle it just produced. It would need a trusted root for the target deployment, which this package cannot scope to a caller's custom `FulcioURL`/`RekorURL` without a TUF fetch. Verification stays the caller's step through `container/verifier`.
- The fixture identity token is unsigned on purpose. Real Fulcio verifies the token against its issuer; sigstore-go only base64-decodes the claims to recover the subject. Signing the fixture would imply a check that does not happen here.

### DIFF
--- a/container/signer/cosign_attach.go
+++ b/container/signer/cosign_attach.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"crypto"
+	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
 	"encoding/hex"
@@ -30,7 +31,11 @@ import (
 	protobundle "github.com/sigstore/protobuf-specs/gen/pb-go/bundle/v1"
 	protorekor "github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1"
 	fulciocert "github.com/sigstore/sigstore-go/pkg/fulcio/certificate"
+	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/verify"
 	"github.com/sigstore/sigstore/pkg/signature"
+
+	"github.com/stacklok/toolhive-core/container/verifier"
 )
 
 const (
@@ -141,6 +146,7 @@ func attachCosignSignature(
 	payload []byte,
 	pb *protobundle.Bundle,
 	pub crypto.PublicKey,
+	tm root.TrustedMaterial,
 ) (bool, error) {
 	ref, err := name.ParseReference(imageRef)
 	if err != nil {
@@ -177,7 +183,7 @@ func attachCosignSignature(
 		return false, err
 	}
 
-	already, err := alreadySigned(base, payload, pub, cert)
+	already, err := alreadySigned(ctx, keychain, imageRef, digestStr, base, payload, pub, cert, tm)
 	if err != nil {
 		return false, err
 	}
@@ -208,14 +214,18 @@ func attachCosignSignature(
 	return true, nil
 }
 
-// alreadySigned reports whether base already carries a signature layer
-// equivalent to the one about to be attached, dispatching to the dedupe
-// check that matches the signature's layout: identity comparison for a
-// certificate-bearing (keyless) signature, key comparison otherwise. See
-// signedByIdentity and signedByKey for why these must be different checks.
-func alreadySigned(base v1.Image, payload []byte, pub crypto.PublicKey, cert *certMaterial) (bool, error) {
+// alreadySigned reports whether the artifact already carries a signature
+// layer equivalent to the one about to be attached, dispatching to the
+// dedupe check that matches the signature's layout: a chain-verified
+// identity check for a certificate-bearing (keyless) signature, key
+// comparison against base's raw layers otherwise. See keylessAlreadySigned
+// and signedByKey for why these must be different checks.
+func alreadySigned(
+	ctx context.Context, keychain authn.Keychain, imageRef, digestStr string,
+	base v1.Image, payload []byte, pub crypto.PublicKey, cert *certMaterial, tm root.TrustedMaterial,
+) (bool, error) {
 	if cert != nil {
-		return signedByIdentity(base, payload, cert.summary)
+		return keylessAlreadySigned(ctx, keychain, imageRef, digestStr, payload, cert, tm)
 	}
 	return signedByKey(base, payload, pub)
 }
@@ -310,78 +320,87 @@ func signedByKey(img v1.Image, payload []byte, pub crypto.PublicKey) (bool, erro
 	return false, nil
 }
 
-// signedByIdentity reports whether img already carries a certificate-bearing
-// signature layer from the same signer identity as summary that ALSO
-// cryptographically verifies against payload — i.e. whether this keyless
-// signer has already produced a valid signature over this exact artifact.
-// Identity match alone is not enough to dedupe on: a layer whose signature
-// does not verify against its own certificate is not "already signed",
-// merely damaged, and treating it as equivalent would let a corrupted or
-// mismatched existing layer permanently block a legitimate re-sign from
-// ever attaching a valid signature. This is the keyless counterpart to
-// signedByKey: comparing certificate bytes, or the signature itself, would
-// never dedupe, because keyless signing mints a fresh ephemeral keypair and
-// certificate on every run. Identity — certificate SAN plus OIDC issuer —
-// is what stays stable across runs, and is also what a verifier's trust
-// policy binds to (see container/verifier's identityPolicyOption), so it is
-// the right notion of "already signed by this signer", checked alongside —
-// never instead of — the cryptographic check.
-func signedByIdentity(img v1.Image, payload []byte, summary fulciocert.Summary) (bool, error) {
-	manifest, err := img.Manifest()
-	if err != nil {
-		return false, fmt.Errorf("reading signature manifest layers: %w", err)
+// keylessAlreadySigned reports whether the artifact at imageRef/digestStr
+// already carries a genuinely trusted keyless signature from cert's
+// identity over payload. See keylessLayerTrusted for why identity match
+// alone (certificate SAN plus OIDC issuer — the keyless counterpart to
+// signedByKey's public-key comparison, since keyless signing mints a fresh
+// ephemeral keypair and certificate on every run) is NOT enough to dedupe
+// on: unlike signedByKey's pub, which the caller supplies as ground truth,
+// the SAN and issuer extensions are attacker-visible fields inside the
+// certificate itself, so a self-signed certificate can claim any identity —
+// this must additionally chain-verify.
+//
+// This re-reads the registry through verifier.RetrieveBundles rather than
+// scanning the existing signature manifest's raw layers directly (as the
+// key-signed path does): only RetrieveBundles' reconstruction carries the
+// certificate and transparency-log material coreverifier.VerifyBundle
+// needs, and rebuilding that here would duplicate container/verifier's own
+// read side. A failure here — including "nothing attached yet" — is
+// treated as not-yet-signed rather than propagated: at worst this appends a
+// redundant layer, which append-never-replace already makes safe, and that
+// is a better failure mode than blocking a legitimate sign over a
+// transient read problem.
+func keylessAlreadySigned(
+	ctx context.Context, keychain authn.Keychain, imageRef, digestStr string,
+	payload []byte, cert *certMaterial, tm root.TrustedMaterial,
+) (bool, error) {
+	full := imageRef
+	if !strings.Contains(imageRef, "@") {
+		full = imageRef + "@" + digestStr
 	}
-	for _, l := range manifest.Layers {
-		pemCert := l.Annotations[annotationCosignCertificate]
-		if pemCert == "" {
-			continue
-		}
-		existing, err := parseLeafCertificate([]byte(pemCert))
-		if err != nil {
-			// A layer we cannot parse is not one of ours; leave it alone
-			// rather than failing the whole push over someone else's
-			// malformed annotation.
-			continue
-		}
-		existingSummary, err := fulciocert.SummarizeCertificate(existing)
-		if err != nil {
-			continue
-		}
-		if existingSummary.SubjectAlternativeName != summary.SubjectAlternativeName ||
-			existingSummary.Issuer != summary.Issuer {
-			continue
-		}
-		if verifies, _ := certLayerVerifies(l, existing, payload); verifies {
+	bundles, err := verifier.RetrieveBundles(ctx, full, keychain)
+	if err != nil {
+		return false, nil //nolint:nilerr // see doc comment: a read failure here means "append", not "fail the sign"
+	}
+	opts, err := verifier.DefaultVerifierOptions()
+	if err != nil {
+		return false, nil //nolint:nilerr // embedded verifier options; failure here is not expected in practice
+	}
+	for _, b := range bundles {
+		if b.HasCertificate() && keylessLayerTrusted(b, tm, opts, payload, cert.summary) {
 			return true, nil
 		}
-		// Same identity, but this layer's signature does not verify against
-		// its own certificate and payload — it is corrupt or stale, not a
-		// prior valid signing. Fall through rather than dedupe on it, so a
-		// new, valid signature still gets appended.
 	}
 	return false, nil
 }
 
-// certLayerVerifies reports whether layer's attached signature annotation
-// cryptographically verifies against cert's public key and payload. A
-// decode or verifier-construction failure is reported as "does not verify"
-// rather than propagated: the annotation is registry-supplied data, and the
-// caller's fallback (treat as not-yet-signed, append a new layer) is
-// already the correct response to a layer that fails to check out.
-func certLayerVerifies(layer v1.Descriptor, cert *x509.Certificate, payload []byte) (bool, error) {
-	encoded := layer.Annotations[annotationCosignSignature]
-	if encoded == "" {
-		return false, errors.New("layer carries a certificate but no signature annotation")
+// keylessLayerTrusted reports whether bundle b is a genuine, chain-verified
+// Fulcio/Rekor signature from summary's identity over payload — not merely
+// a certificate carrying a matching SAN/issuer extension. Both fields are
+// public, unauthenticated data embedded in the certificate: anyone with
+// registry write access can mint a self-signed certificate carrying a
+// victim's SAN and OIDC-issuer extension, sign the known simple-signing
+// payload with the matching (self-generated) key, and push it as a
+// signature layer. A check that only confirms internal self-consistency
+// (the signature verifies against ITS OWN certificate's key) would accept
+// that forgery as "already signed by this identity" and silently skip
+// attaching the real, Fulcio-backed signature — a signing-pipeline bypass,
+// not a forged verification (a real consumer's install/verify step still
+// rejects the self-signed certificate on chain-of-trust), but a bypass
+// nonetheless. This instead runs the exact chain-of-trust and
+// transparency-log verification a real consumer's install/verify step
+// applies — container/verifier's embedded trust root and
+// DefaultVerifierOptions — so only a genuinely Fulcio-issued, Rekor-logged
+// certificate can dedupe.
+//
+// The digest check runs before the (comparatively expensive) chain
+// verification: a chain-valid signature over some OTHER payload from this
+// same identity says nothing about whether THIS payload has already been
+// signed, and checking it first also avoids wasted verification work.
+func keylessLayerTrusted(
+	b verifier.Bundle, tm root.TrustedMaterial, opts []verify.VerifierOption, payload []byte, summary fulciocert.Summary,
+) bool {
+	sum := sha256.Sum256(payload)
+	if b.DigestAlgo != "sha256" || b.DigestHex != hex.EncodeToString(sum[:]) {
+		return false
 	}
-	raw, err := base64.StdEncoding.DecodeString(encoded)
-	if err != nil {
-		return false, fmt.Errorf("decoding signature annotation: %w", err)
+	vr, err := verifier.VerifyBundle(b, tm, nil, opts...)
+	if err != nil || vr.Signature == nil || vr.Signature.Certificate == nil {
+		return false
 	}
-	sigVerifier, err := signature.LoadVerifier(cert.PublicKey, crypto.SHA256)
-	if err != nil {
-		return false, fmt.Errorf("loading verifier for existing certificate: %w", err)
-	}
-	return sigVerifier.VerifySignature(bytes.NewReader(raw), bytes.NewReader(payload)) == nil, nil
+	return vr.Signature.Certificate.SubjectAlternativeName == summary.SubjectAlternativeName &&
+		vr.Signature.Certificate.Issuer == summary.Issuer
 }
 
 // certMaterial is the certificate and transparency-log evidence for a
--- a/container/signer/cosign_attach.go
+++ b/container/signer/cosign_attach.go
@@ -127,6 +127,13 @@ func parseManifestDigest(digestStr string) (digest.Digest, error) {
 // pub is the public key used for the key-signed dedupe check below — it is
 // separate from pb because the key-signed bundle itself carries no key
 // material, only a hint.
+//
+// The returned bool reports whether a new layer was actually appended:
+// false means the artifact was already signed by this identity/key and
+// attachCosignSignature was a no-op. This matters to the caller — pb's
+// signature is randomised per call, so when nothing was appended, pb does
+// NOT represent what is now on the registry; only the pre-existing layer
+// does. See SignOCI's use of this to decide what to return as Result.Bundle.
 func attachCosignSignature(
 	ctx context.Context,
 	keychain authn.Keychain,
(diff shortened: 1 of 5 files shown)


rdimitrov (dismissed):

JAORMX (commented):
Panel review (spec vs stacklok/toolhive#6307, repo standards, six specialist reviewers — every technical claim verified against the pinned deps on disk: sigstore-go v1.3.0, sigstore v1.10.9, rekor v1.5.3).

**Spec axis** — clean within this PR's claimed scope. Keyless flow, `ErrAmbiguousSigningMethod`, signing-time defaulting, mandatory Rekor v1 entry (verified: `container/verifier` requires SCT=1/tlog=1/observer=1), unchanged key path — all confirmed against the diff. Still open against #6307 (correctly deferred, not done here): CLI/push-API credential plumbing, Sigstore-staging E2E, stock `cosign verify` interop.

**Standards** — CLAUDE.md Boundaries says never modify `go.mod` dependencies without being asked; the 5 indirect→direct promotions are test-only and correct (`go.sum` unchanged), but please confirm an approver is fine with them. Also: table-driven tests for new functions — `TestKeylessBundleOptionsDefaultToPublicGood` is the clear candidate.

**Domain** — the likely blocking finding: keyless dedupe verifies the layer's signature against the key in its *own* cert, never chaining that cert to the configured Fulcio root (inline on cosign_attach.go). A registry writer can preload a self-signed same-identity layer, making signing a silent no-op that returns attacker material as success. Also inline: no HTTPS/loopback validation on FulcioURL/RekorURL (bearer token; `networking.ValidateEndpointURL` exists); Fulcio-vs-Rekor failure stage unidentifiable in errors; goroutine+recover wrapper scoped wider than the two verified upstream defects (`pkg/sign/certificate.go:175-207`, `pkg/tle/tle.go:33-44` — claims confirmed).

**Verified non-findings**: hardcoded-endpoint rationale holds (`root.CertificateAuthority` exposes only `Verify`); Rekor-v1-only holds (v2 needs a TSA per `pkg/sign/signer.go:171-189`); no `IdentityToken` leakage into local errors; unsigned-JWT parsing is test-only; v0.3/v0.1 `Result.Bundle` shape variance round-trips through `container/verifier` (pinned by `TestSignOCIKeylessRoundTrip`).

container/signer/signer.go
```diff
@@ -105,34 +204,269 @@ func (d *Default) SignOCI(ctx context.Context, ref, digestStr string, opts Optio
...
+		CertificateProvider: sign.NewFulcio(&sign.FulcioOptions{
+			BaseURL: fulcioURL,
+			Retries: keylessRetries,
+		}),
+		CertificateProviderOptions: &sign.CertificateProviderOptions{IDToken: opts.IdentityToken},
+		TransparencyLogs: []sign.Transparency{
+			sign.NewRekor(&sign.RekorOptions{
+				BaseURL: rekorURL,
+				Retries: keylessRetries,
+				Version: rekorAPIVersionV1,
```
JAORMX, line 572: **[High — cross-confirmed] No scheme validation before the bearer token goes out.** `opts.FulcioURL` flows into `sign.NewFulcio` unchecked; sigstore-go v1.3.0 attaches `Authorization: Bearer <IdentityToken>` at `pkg/sign/certificate.go:178-187` for any URL — so `FulcioURL: "http://…"` (non-loopback) sends the OIDC token in the clear (CWE-319 / OWASP A02:2021). The caveat is documented on `IdentityToken`'s godoc, but someone configuring the URL reads *that* field's docs, which don't mention it.

Related (verified): the defaults apply independently, so setting only `FulcioURL` pairs a private Fulcio with public Rekor, producing a bundle that verifies under no standard trust root — nothing rejects or documents that pairing.

Suggest validating both endpoints here after defaulting — the repo already has `networking.ValidateEndpointURL` (`networking/utilities.go:186`) — requiring HTTPS except literal loopback (which your tests need). `keylessBundleOptions` would return an error; a table-driven test could cover http/file/missing-host vs HTTPS/loopback.

container/signer/cosign_attach.go
```diff
@@ -191,14 +198,14 @@ func attachCosignSignature(
...
+		return false, fmt.Errorf("building signature manifest: %w", err)
 	}
 	img = mutate.MediaType(img, types.OCIManifestSchema1)
 
 	if err := remote.Write(sigTag, img, remoteOpts...); err != nil {
-		return fmt.Errorf("pushing signature manifest: %w", err)
+		return false, fmt.Errorf("pushing signature manifest: %w", err)
 	}
-	return nil
+	return true, nil
```
JAORMX, line 214: **[High — likely the blocking one] The keyless dedupe path (just above, in `signedByIdentity`) verifies the layer's signature against the key inside the layer's *own* certificate — but nothing checks that certificate chains to the configured Fulcio root (CWE-345).** A registry writer can mint a self-signed cert carrying the victim's SAN and OIDC-issuer extension (both public by design), sign the known simple-signing payload, and push it as a `.sig` layer. Dedupe then accepts it, attach is skipped, and `bundleMatchesSigner` (signer.go:373-404) repeats the same untrusted checks to return the attacker's bundle as a successful `Result`. Downstream verification still rejects the self-signed cert, so it isn't a forgery — but `SignOCI` reports success while the legitimate Fulcio-backed signature is silently never attached (signing-pipeline bypass / DoS).

Options: verify a candidate's full bundle against trusted material before accepting the no-op, or conservatively always append on the keyless path. Test: preload the tag with a self-signed same-identity layer that signature-verifies, assert SignOCI still appends.

container/signer/signer.go
```diff
@@ -105,34 +204,269 @@ func (d *Default) SignOCI(ctx context.Context, ref, digestStr string, opts Optio
...
+		fulcioURL = DefaultFulcioURL
+	}
+	rekorURL := opts.RekorURL
+	if rekorURL == "" {
+		rekorURL = DefaultRekorURL
+	}
+	return sign.BundleOptions{
+		Context: ctx,
+		CertificateProvider: sign.NewFulcio(&sign.FulcioOptions{
+			BaseURL: fulcioURL,
```
JAORMX, line 564: **[Medium — cross-confirmed] This wrapper is scoped wider than the two verified upstream defects.** I checked both claims and they're accurate: sigstore-go's Fulcio client builds its request with `http.NewRequest` (caller ctx never attached; default 30s client timeout — `pkg/sign/certificate.go:175-207`), and rekor nil-derefs `Verification.InclusionProof` (`pkg/tle/tle.go:33-44,88-96`). But this now wraps *every* `sign.Bundle` call, including file-key signing, so an unrelated panic in payload signing / keypair code / bundle assembly is reclassified as a Fulcio-or-Rekor network error with the stack trace discarded. And the ctx race abandons the Fulcio request until its own timeout rather than fixing cancellation.

Suggest narrowing: a RoundTripper that clones each request onto the caller's context for the Fulcio transport; panic-recover only around the Rekor adapter; apply the wrapper only on the keyless path; and file both defects upstream so this can be deleted when fixed.

container/signer/signer.go
```diff
@@ -105,34 +204,269 @@ func (d *Default) SignOCI(ctx context.Context, ref, digestStr string, opts Optio
...
+			Retries: keylessRetries,
+		}),
+		CertificateProviderOptions: &sign.CertificateProviderOptions{IDToken: opts.IdentityToken},
+		TransparencyLogs: []sign.Transparency{
+			sign.NewRekor(&sign.RekorOptions{
+				BaseURL: rekorURL,
+				Retries: keylessRetries,
```
JAORMX, line 571: **[High — cross-confirmed] The Fulcio-vs-Rekor failure stage is unidentifiable from these errors.** `"building sigstore bundle: %w"` only relays sigstore-go's text (unstable dependency wording), and the recover path says "a Fulcio or Rekor response could not be processed" — an operator can't distinguish token-rejected, Fulcio-unreachable, Rekor-down, or malformed-response without guessing, and a consumer can't branch programmatically on the stage. Suggest wrapping with stage context (`requesting signing certificate from Fulcio (%s): %w` / `submitting to Rekor (%s): %w`) — and if applicable, exported stage sentinels or a typed operation error so `errors.Is`/`errors.As` works.

container/signer/signer.go
```diff
@@ -1,49 +1,145 @@
...
+	// flow: a single-use ephemeral key pair is minted, Fulcio issues a
+	// short-lived certificate binding it to the token's identity, and the
+	// signature is submitted to Rekor. Obtaining the token is the caller's
+	// responsibility — see the package doc.
+	//
+	// This token is sent as a bearer credential to FulcioURL. A caller
+	// setting FulcioURL to anything other than an HTTPS (or loopback, for
+	// tests) endpoint sends it in the clear.
+	IdentityToken string
+	// FulcioURL overrides the certificate authority for keyless signing.
```
JAORMX: **[Medium] "zero Options value targets the public-good deployment" misstates the zero-value contract** — bare `Options{}` fails with `ErrKeyRequired`; public-good defaulting only happens when `IdentityToken` is set and the URLs are empty. Suggest: "Empty means DefaultFulcioURL when keyless signing is selected" (same for RekorURL below). Also worth stating on SignOCI/Options that keyless performs outbound network egress to public-good services by default — a posture change from "no key ⇒ always error" that air-gapped consumers should see before runtime.

container/signer/signer.go
```diff
@@ -105,34 +204,269 @@ func (d *Default) SignOCI(ctx context.Context, ref, digestStr string, opts Optio
...
+	bundles, err := verifier.RetrieveBundles(ctx, full, keychain)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving previously attached signature: %w", err)
+	}
+	cert, err := certMaterialFromBundle(pb)
+	if err != nil {
+		return nil, err
+	}
+	for _, b := range bundles {
+		if bundleMatchesSigner(b, payload, cert, pub) {
```
JAORMX: **[Medium] The dedupe round-trip could return the matched bundle instead of rediscovering it.** `attachCosignSignature` found the matching layer and verified it, but returns only a bool; this path then re-reads the registry through `verifier.RetrieveBundles` and re-runs identity+signature matching below. Two costs: a second registry read against state that may have changed since attach's read, and signer now imports `container/verifier` in production code — coupling write-side to the full verification discovery API (a verifier-side discovery change can break signing). Prefer returning the matched layer/bundle (or descriptor) from the attach side; if a shared codec for the cosign annotations is needed between the two packages, keep it narrow and internal. Layering discussion — flagging as judgement, not mechanical.

container/signer/signer.go
```diff
@@ -105,34 +204,269 @@ func (d *Default) SignOCI(ctx context.Context, ref, digestStr string, opts Optio
...
+		if err != nil {
+			return false
+		}
+		return sigVerifier.VerifySignature(bytes.NewReader(msgSig.GetSignature()), bytes.NewReader(payload)) == nil
+	}
+	sigVerifier, err := signature.LoadVerifier(pub, crypto.SHA256)
+	if err != nil {
+		return false
+	}
+	return sigVerifier.VerifySignature(bytes.NewReader(msgSig.GetSignature()), bytes.NewReader(payload)) == nil
```
JAORMX: **[Low] Identity-match + cert-signature verification now lives in two places** — here in `bundleMatchesSigner` and in `signedByIdentity` (`cosign_attach.go`). They must stay in lockstep (the comment above this func says it "mirrors" the attach side) — one shared helper would make the next change to dedupe semantics single-site. Low only because it's two sites today (Rule of Three).

container/signer/keyless_test.go
```diff
@@ -0,0 +1,1076 @@
...
+	require.NoError(t, err)
+	now := time.Now()
+	rootTmpl := &x509.Certificate{
+		SerialNumber:          big.NewInt(1),
+		Subject:               pkix.Name{CommonName: "toolhive-test-sigstore-root"},
+		NotBefore:             now.Add(-time.Hour),
+		NotAfter:              now.Add(time.Hour),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+		IsCA:                  true,
```
JAORMX, line 115: **[Info] Verified reuse opportunity:** the pinned `sigstore-go@v1.3.0` ships `pkg/testing/ca.GenerateRootCa()` and `GenerateFulcioIntermediate` (`pkg/testing/ca/ca.go:814,841`) — the ~40-line root/intermediate fixture in `newTestSigstore` could delegate to those. Separately, `claimsFromToken`'s manual split/base64/JSON decode could be `jwt.NewParser().ParseUnverified` with `jwt.RegisteredClaims` (golang-jwt/v5 is already a direct dep). Test-only, take or leave.

container/signer/signer.go
samuv: Fixed: `keylessBundleOptions` now validates both `FulcioURL` and `RekorURL` with `networking.ValidateEndpointURL` (HTTPS required except loopback) before either is used to build `sign.BundleOptions`, so a non-loopback plain-HTTP endpoint is rejected before the bearer token is ever attached to a request. Regression test: `TestKeylessBundleOptionsRejectsInsecureEndpoint`. Also moved the caveat onto `FulcioURL`/`RekorURL`'s own doc comments instead of leaving it only on `IdentityToken`'s.

container/signer/cosign_attach.go
samuv: Fixed: dedupe for cert-bearing layers now verifies the full chain of trust via `verifier.VerifyBundle`/`root.TrustedMaterial` before comparing SAN+issuer, rather than trusting a self-consistent signature over the layer's own (attacker-controlled) certificate. New shared helper `keylessLayerTrusted` (used by both attach's dedupe and `bundleMatchesSigner`) does: (1) confirm digest match, (2) run full bundle verification against the configured trust material, (3) only then compare the verified certificate's SAN/issuer. A self-signed cert with a matching SAN/issuer can no longer pass — `TestAttachCosignSignatureCertDedupeRequiresChainOfTrust` is the regression test (mutation-verified: reverting the chain check to always-true makes it fail). In production this verifies against `verifier.OfflineTrustedMaterial()` (the embedded production Sigstore root); tests inject an in-process CA as trust material via a new `newDefaultForTest` constructor, since they can't chain to the real production root.

container/signer/signer.go
samuv: Fixed: split `signBundle` into a synchronous key-signed path (no goroutine, no recover — a panic there is this package's own bug and should propagate as one) and a keyless-only `keylessSignBundle` that keeps the goroutine+recover+ctx-race workaround scoped to the actual untrusted-network-response problem it exists for.

container/signer/signer.go
samuv: Fixed: `keylessSignBundle` now calls `opts.CertificateProvider.GetCertificate` directly first, tagging any failure as "requesting signing certificate from Fulcio: %w", then hands `sign.Bundle` the already-fetched cert via a `cachedCertificate` provider so its internal call is a cache hit rather than a second Fulcio round trip — leaving Rekor submission as the only remaining fallible step, tagged "submitting to Rekor: %w". Verified safe by reading `sign.Bundle`'s source for this project's specific `BundleOptions` (no `TimestampAuthorities`, no `TrustedRoot`) to confirm no other fallible step exists between cert-acquisition and Rekor submission. `TestSignOCIKeylessSurfacesFulcioFailure` and the new `TestSignOCIKeylessSurfacesRekorFailure` each assert the error names the correct stage and not the other.

container/signer/signer.go
samuv: Fixed the doc comment — it now says empty `FulcioURL`/`RekorURL` default to the public-good instances only when keyless signing is selected (`IdentityToken` set), not that a bare zero-value `Options{}` targets them (that still fails with `ErrKeyRequired`). Also added the outbound-egress-by-default note.

container/signer/signer.go
samuv: Left this as-is for now — happy to take it as a follow-up if you'd rather have it in this PR. Two things about it: `pb`'s signature can't be reused directly regardless of what attach returns, since every signing operation is randomized (fresh ephemeral key + Fulcio cert for keyless, ECDSA's own nonce for key re-sign) — that's the existing comment on `resultBundleJSON`. What your suggestion would actually save is the second `verifier.RetrieveBundles` registry read: `keylessAlreadySigned` already walks the bundles looking for a match, so it could return the matched `verifier.Bundle` instead of a bool, and `previouslyAttachedBundleJSON` could reuse it instead of re-listing. I didn't make that change here to keep this pass scoped to the chain-of-trust fix and the other findings above; the `container/verifier` import stays either way since dedupe itself needs to verify against it.

container/signer/signer.go
samuv: This one's resolved as a side effect of the chain-of-trust fix: `keylessLayerTrusted` is now the single shared helper for cert-bearing identity+chain verification, called from both `keylessAlreadySigned` (cosign_attach.go) and `bundleMatchesSigner` (signer.go). The two call sites still exist because they're matching against different things (attach matches against the registry's existing layers; the dedupe-result path re-derives the same match after attach reports a no-op), but the verification logic itself is single-site now.

container/signer/keyless_test.go
samuv: Left as-is — noted, and it's a legitimate simplification, but swapping the fixture for `ca.GenerateRootCa()`/`jwt.NewParser` in the same pass as the chain-of-trust rework risked masking whether a test failure came from the fixture change or the actual fix. Fine as a separate cleanup follow-up if you'd like it filed as an issue.

samuv:
Addressed the domain findings from the panel review (replies inline on each):

- **Keyless dedupe now verifies chain of trust** (the likely-blocking finding) instead of trusting SAN+issuer over a self-consistent signature — a self-signed same-identity cert can no longer be accepted as an existing signature.
- **FulcioURL/RekorURL scheme validation** via the shared `networking.ValidateEndpointURL` (HTTPS required except loopback), so the bearer token never goes out over plain HTTP.
- **Fulcio vs. Rekor failure stage is now identifiable** in errors (cert-acquisition is fetched directly and cached into `sign.Bundle`, isolating Rekor submission as the only remaining fallible step).
- **recover()/ctx-race wrapper narrowed** to the keyless path only — the key-signed path is now a plain synchronous call.
- Corrected the `Options.FulcioURL`/`IdentityToken` doc comments' zero-value contract.

On the two "Standards" notes in the review body:
- The 5 indirect→direct `go.mod` promotions are unchanged by this round of fixes — still test-only (driven by `keyless_test.go`'s direct imports), `go.sum` is untouched. Flagging again for an approver's explicit sign-off per CLAUDE.md's dependency-change boundary, since I can't self-approve that.
- Didn't convert `TestKeylessBundleOptionsDefaultToPublicGood` to table-driven — it asserts several distinct properties of one configuration rather than varying inputs, so a table wouldn't reduce duplication there. `TestKeylessBundleOptionsRejectsInsecureEndpoint` (new) does exercise multiple input variations sequentially; happy to make that one tabular if you'd prefer.

Also note: the branch was rebased onto main after the review to pick up the `golang.org/x/mod` CVE fix (#236), so the diff base moved — the content at the review's commit (`697e3d4`) and this branch's current pre-fix commit are identical, only the fixes above are new.

Verification: `go build`/`go vet` clean, `task lint` 0 issues, `container/signer` full suite with `-race` green, and `task test` (whole module) green.

jhrozek (approved):

JAORMX (approved):

